PR #444, opened 2014-12-29
Bring blogging points to the front

konklone:
Link to the blogging guide more forcefully from the README, and add some high level bullets to the blogging guide about metadata and workflow.


konklone:
This is all docs, and non-controversial, so I'm going to self-merge.

PR #457, opened 2015-01-09
dedupe from internal blogging guide

afeld:


konklone:
Thanks @afeld - made a tiny tweak to the asset line, looks good to me.

PR #458, opened 2015-01-13
Final edits before publishing

gboone:
Post is live on staging, this PR is for any pre-publishing edits like capitalization. It also adds `future: false` to the `_config.yml` to hide any posts with publish dates that are in the future.


konklone:
:+1:

PR #460, opened 2015-01-13
Add Moncef's photo and name to homepage.

monfresh:


konklone:
Nice! :+1: 

@gboone I haven't totally kept track of the data-shuffling that recently happened with the 18F Hub, but I assume we can merge this in here and then update in the Hub's data as necessary for the time being?

@monfresh This is just fine, but for further PRs, branch PRs are more helpful than fork PRs (since we can hop on the branch and collaborate before merging).

monfresh:
@konklone Absolutely. Will do from now on now that I have access to the 18F org.

gboone:
@konklone @monfresh for now (but hopefully not much longer) you have to add yourself here and [in the `data-private` repo](https://github.com/18F/data-private) to update the [Hub](https://18f.gsa.gov/hub).

PR #468, opened 2015-01-18
Try out Travis' new container infrastructure

konklone:
Uses Docker and is way faster:

http://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: ruby
 rvm:
   - 2.1.2
 
+sudo: false
+
 install: script/bootstrap
 script: script/cibuild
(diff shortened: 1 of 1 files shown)


konklone:
Well, I don't have the data to say whether this was actually faster. But, er, it seemed fast enough, and it worked. 

https://travis-ci.org/18F/18f.gsa.gov/builds/47467663

Seems like it's probably Travis' future. What do you think?

gboone:
Yeah, not sure if it's faster but I'm generally pro- The Future. Seems like a step in the right direction, anyway.

PR #469, opened 2015-01-19
Update DAP snippet - new location, upgraded version

konklone:
This updates the DAP snippet on the 18f.gsa.gov website. It does three things:

* Stop self-hosting the snippet, and point to a cloud-hosted location.
* Uses the new `analytics.usa.gov` domain for the DAP snippet. I've uploaded v1.0.1 [as found in the official repo](https://github.com/GSA/DAP-Gov-wide-GA-Code/blob/master/Universal-Federated-Analytics.1.0.js) (after [minifying and compressing it](https://github.com/GSA/analytics.usa.gov/blob/b0905dd05338cb55aab37e1685122978166742dd/dap.md#minifying)).
* Updates to Universal Analytics. We've so far been using the old deprecated snippet, which I didn't know.

cc @gbinal @tdlowden


gboone:
This looks good.

PR #399, opened 2014-11-27
A first pass at a third party disclosure page

konklone:
I added a page for third party service disclosure, and attempted to explain how each one plays into our overall service. It also explains what extra things we do to try to protect privacy.

You can see the page, and the new link in the footer, in this screenshot (updated 2014-11-03):

18f digital services delivery

If we keep going with our current plan, we can eventually remove the webfont dependency, and limit OCSP privacy leakage to browsers on Safari (which doesn't support stapling at all, but does make OCSP calls). (Also, we'll be switching from Symantec/Verisign to Comodo, when we get our new cert...someday.)

This [giant Symantec PDF](http://www.symantec.com/content/en/us/about/media/repository/stn-cps.pdf) shows Symantec referencing their overall privacy policy to cover their work as a CA.

Fixes #293.

PR #474, opened 2015-01-21
Revised privacy language

gboone:
It came to our attention while discussing #399 that the privacy language we have on 18f.gsa.gov is inaccurate and not in keeping with terms we are using on other 18F projects. This commit integrates more accurate terms suggested by @polastre.

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -122,7 +122,7 @@ permalink: /
     </div>
     <div class="contact-words">
       <p>Leave us your name and email if you'd like more information about the way we work.</p>
-      <p class="small">Your privacy and security are important to us, and we'll never share your information. Please see <a href="http://www.gsa.gov/portal/content/116609">GSA's Privacy and Security Notice</a> for more information.</p>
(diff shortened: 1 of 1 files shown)


NoahKunin:
:+1: to this, thanks

polastre:
:shipit:

NoahKunin:
For the (public) record, this was a mistake on my part when reviewing copy. In the context of "sharing data with _third_ parties" the statement was accurate. It's in the context of under the law sharing data with other teams within government it was inaccurate.

konklone:
And to disambiguate - I think what you mean, @NoahKunin, is how the statement works in relation to the name and email address we ask people to put into the newsletter signup form. We share it with other GSA teams, but not with third parties. (Though if we switch from HYDRA to e.g. Mailchimp, that would change.)

I originally brought this statement up in #399 in the context that it was the only language relating to privacy on our whole website. So if it were interpreted as a statement of how we handle _all_ of our visitors' information (including their user agent, request metadata, etc.), it wouldn't be accurate, since Google and Amazon both see that information.

So I'm a :+1: for merging this, for sure, but it doesn't address the full scope of how our website handles visitor information.

NoahKunin:
@konklone Correct, this is a technical hotfix while we tracker the larger issue.

PR #475, opened 2015-01-21
Join us post

gboone:


konklone:
Doing a couple quick technical edits, then will merge.

konklone:
Is there an OG image?

konklone:
I changed the images to move them to the `assets/blog` dir.

I'm done - after the OG image link is added, :+1: any time.

gboone:
Good catch on those Google encoded URLs, @konklone, I pushed up the og-image with 5898875

PR #479, opened 2015-01-22
Deploy temporary email change

gboone:
We can revert this if necessary.


NoahKunin:
Ok @gboone @konklone, I've fixed the real email, so you can now revert.

Issue #486, opened 2015-01-26
Refined author display


gboone:
Currently we're using a filter that pulls in the author's full name from the `team.yml` file. This works great on blog posts, but less elegantly next to post excerpts in tag archives and on the home page.

Related to #441

PR #489, opened 2015-01-27
Pulls data from the hub's data files

gboone:
Adds:
- the Gemfile,
- Gemfile.lock,
- _data/private submodule
- a go script!

This isn't perfect, so I've also updated the site index to filter out some data in the mean time. Might be worth taking on the time to take on #486 here and just use an author filter to figure out whether to display
someone's face.

Other problems: some team members have incomplete data (no first_name or last_name) and others have no photo. We'll have to take on all that before merging.


mbland:
Looks like you need to add the following to `.travis.yml`:

```yaml
git:
  submodules: false
```

go
```diff
@@ -0,0 +1,122 @@
...
+# Date:   2015-01-10
+
+MIN_VERSION = "2.1.5"
+
+unless RUBY_VERSION >= MIN_VERSION
+  puts <<EOF
+
+*** ABORTING: Unsupported Ruby version ***
+
+Ruby version #{MIN_VERSION} or greater is required to work with the Hub, but
```
mbland: s/Hub/Dashboard/g

gboone:
Alright. Finally got Travis sorted out. Going to work on migrating the new post script into the go script.

Issue #455, opened 2015-01-08
Bring sub-features into one repo


meiqimichelle:
For discussion:

There are now several repos that are actually a part of 18f.gsa.gov and it is getting more and more difficult to keep assets in sync across them (basically a copy-and-paste job at the moment).

@mbland and @afeld and I started thinking about how to solve this, and then realized that we might have moved past the need for those projects to be in separate repos at all.

Can we put them all under the same roof and treat them as they are -- one cohesive offering?

@mbland also suggested that we can still maintain a template-style, forkable /dashboard or hub (or any other part) as separate repos. The 18f.gsa.gov implementation would then serve as the living example of the templates.

konklone:
> Can we put them all under the same roof and treat them as they are -- one cohesive offering?

It's a lot easier operationally, in most ways. One way where it's not is that it'll be quite a busy issue tracker. The 18f.gsa.gov repo so far has had 160 issues and almost 300 pull requests, just for the landing page and blog -- and it's about to undergo major surgery as part of the redesign effort. That's something to plan for if we merged them.

> @mbland also suggested that we can still maintain a template-style, forkable /dashboard or hub (or any other part) as separate repos. The 18f.gsa.gov implementation would then serve as the living example of the templates.

If I'm understanding that right, there'd be 4 repos? One is just the template, and then 18f.gsa.gov, hub, and dashboard would all pull from it?

mbland:
@konklone Three repos, the other way around: Dashboard and Hub repos would be templates, perhaps.

mbland:
@konklone Another thing to think about: Most of the data will be sourced form data-private, and most of the Hub logic will hopefully be extracted into gems, living in their own repos (like https://github.com/18F/hash-joiner and https://github.com/18F/weekly_snippets). That may cut down on the issue/PR load for 18f.gsa.gov itself. (I think. I hope.)

afeld:
Laid out some of the technical considerations here: https://github.com/18F/18f.gsa.gov/issues/456

gboone:
[...]

PR #492, opened 2015-01-28
Re-apply pulling data from the hub

gboone:
This will reintroduce the changes made by #489 but should not be merged until the following problems are sorted out:

- [x] some team members have incomplete data (no first_name or last_name)
- [x] others have no photo
- [x] README should be updated to point at the go script rather than the `script/` directories


gboone:
@afeld looks like you didn't break the build so, we're good :smile:

gboone:
The latest few commits here add some modifications to the `author` plugin that includes a filter we can use on the home page like this: `{{ member | with_pic }}`. All that's left to merge this is an update to the main README with updated instructions on getting the site running locally and that (for now) we still need you to submit an image to us if you want your face to show up on the front page.

gboone:
Oh, and because we have the `{{ member | with_pic }}` filter we no longer need to check if a team member is hidden before outputting their name and picture. The filter outputs nothing if Jekyll can't find an image with the same `name` as the team member. This should also prevent us from ever having a name show up on the site without a photo again.

gboone:
So, after following along with all the conversation happening around 18F/Hub yesterday I got to thinking about how much it's worth our time merging this and using Jekyll to pull in the Hub's data.

Is there an advantage to doing all this work with JavaScript that requests data from the Hub's API? This PR adds a bit of complexity mostly in the form of git submodules that need to be updated somehow. From what I can tell the possibilities for keeping the data up to date are 

1. The Series of Webhooks plan where `data-private` fires off a webhook to `staging` that updates the data and commits the resulting `team.yml` and `project.yml` files back to this repo. That, in turn triggers the webhook that builds the site.
2. Individuals and the site team periodically refreshing the data by updating the submodule and running `_data/import_data.rb`
3. Any data pulled out of the Hub should be done through the Hub's JSON API.

Advantage to opt. : Ordinary users don't really need to worry about how the data are getting there because it should always be fresh. The disadvantage is that it seems like a lot of automated complexity and that gives me a mild case of the fantods.

Advantage to opt. 2: The deployment infrastructure stays the same. The disadvantage is the relatively higher barrier of entry for folks who don't want to mess around with submodules, ruby, or who really want to build the site as quickly as possible.

Option three seems to cover both the advantages of the other two without either of their disadvantages.

Discuss.

_plugins/author.rb
```diff
@@ -19,6 +19,18 @@ def render(context)
       end
     end
   end
+
+  module AuthorFilter
+    def with_pic(input)
+      if File.exist?("#{Jekyll.sites[0].config['source']}/assets/images/team/#{input['name']}.jpg")
```
mbland: Pretty sure `Jekyll.sites[0].config['source']` can be replaced with `site.source`.

_posts/2014-11-17-taking-control-of-our-website-with-jekyll-and-webhooks.md
```diff
@@ -27,7 +27,7 @@ It may not look a lot different on the surface, but we made some serious infrast
 
 Deploying our blog posts from GitHub instead of managing a separate Tumblr account has drastically simplified our publishing process and opened it up to the whole team. While we still have an approval process for the _content_ of our blog posts, now anyone on the team [can hit the Publish button](https://github.com/18F/18f.gsa.gov/pull/314).
 
-To that end, we've written a [team blogging guide](https://github.com/18F/18f.gsa.gov/blob/staging/blogging.md) to make the blogging process easier for teammates less familiar with Jekyll.
+To that end, we've written a [team blogging guide](https://github.com/18F/18f.gsa.gov/blob/staging/_posts/blogging.md) to make the blogging process easier for teammates less familiar with Jekyll.
```
mbland: This should be `README.md` now, shouldn't it?

mbland:
I support whatever you and the team decide, of course. It does seem like option 3 would suffice, if we've got the JavaScript hands on-deck to help out, which it seems we might. So, basically, the Hub does all the data munging and API publishing, so no one else has to. :-) Then it comes down to using JS to generate the page layouts vs. Jekyll+Liquid+Markdown.

_plugins/author.rb
mbland: Whoops, this is a filter, not a generator plugin; nevermind.

_posts/2014-11-17-taking-control-of-our-website-with-jekyll-and-webhooks.md
gboone: it should!

afeld:
If the plan is still to [combe the sites](https://github.com/18F/18f.gsa.gov/issues/455), probably best to just go with the easiest thing that works for now.

gboone:
> go with the easiest thing that works for now.
In the interest of doing that, let's merge this now.

PR #496, opened 2015-01-29
added meghana

mkhandekar:
picture & name


gboone:
Thanks, @mkhandekar and @konklone

PR #497, opened 2015-01-30
add ozzy

ozzyjohnson:
Just getting around to adding myself.


konklone:
Looks good!

screenshot from 2015-01-29 20 09 35

PR #498, opened 2015-01-30
Takes out the USDS hiring link for now.

gboone:


gboone:
:clap: thanks @konklone

PR #504, opened 2015-02-04
I thought a gradient would be nice here.

Gabrielmtn:
Before I build this site (node, python, and ruby! Ain't life grand?) to edit the _custom.scss and submit a pull request, do other people feel like it's a worthy tweak?

https://drive.google.com/file/d/0ByCk25e3nawEdmZCRkNEd1BzMHc/view?usp=sharing 

(Demo directly injects a -webkit-gradient background to the ".splash h1" class)

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -3,6 +3,7 @@ layout: default
 permalink: /
 ---
 
+<!-- this is the section in question-->
 <section class="splash" style="" role="banner" itemscope itemtype="http://schema.org/WPHeader">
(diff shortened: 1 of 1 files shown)


meiqimichelle:
Hi @Gabrielmtn ! We're in the middle of a site re-info-architecture so we're not changing any of the styles right now. We also have an issue with keeping our `_custom.scss` in sync across several repos that we're working on addressing. No can do on the edit, although thanks for the input!

PR #508, opened 2015-02-05
Adds CORS to the production and staging sites

### LINKED ISSUE #507: Add CORS support
We have a bunch of little JSON endpoints now, on the main site and the hub (and maybe the dashboard?), and we've identified at least one use case for CORS.

This can be tackled centrally in this repo's nginx config, by adding [these lines](https://github.com/konklone/cdns-to-https/blob/gh-pages/nginx/cdn.cors) to the appropriate endpoints:

```
add_header Access-Control-Allow-Headers "X-Requested-With";
add_header Access-Control-Allow-Methods "GET, HEAD, OPTIONS, POST";
add_header Access-Control-Allow-Origin "*";
```

I think we want to remove `POST` from the allowed methods here, especially if we want to just apply it to the whole domain and not think hard about which endpoints need what level of protection.

cc @shawnbot @mbland

konklone:
For the public paths that have content for people to read (aka not the webhooks).

I'd like to test this out on the staging server manually before merging. I'll indicate when that's done.

Fixed #507.

### DIFF
--- a/deploy/18f-site.conf
+++ b/deploy/18f-site.conf
@@ -34,16 +34,19 @@ server {
   location / {
       root   /home/site/production/current/_site;
       default_type text/html;
+      include cors.rules;
   }
 
   location /dashboard {
       root /home/site/production/dashboard/_site;
       default_type text/html;
+      include cors.rules;
   }
 
   location /hub {
       root /home/site/production/hub/_site_public;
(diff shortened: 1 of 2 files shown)


gboone:
@konklone! Thank you this fell off my radar.

PR #510, opened 2015-02-06
Fabfile, go, and import-data scripts updated

gboone:
This PR updates the deployment scripts to take into account the `noexec` setting on the staging/production environments


gboone:
cc: @mbland

deploy/fabfile.py
```diff
@@ -33,26 +33,38 @@
 home = "/home/site"
 log = "%s/hookshot.log" % home
 current = "%s/%s/current" % (home, environment)
+ruby = "/opt/install/rbenv/shims/ruby"
 
-# principal command to run upon update
-command = "cd %s && git pull && bundle exec jekyll build >> %s" % (current, log)
+if environment == 'staging':
+  # on staging, build the site but also push back the extracted data files
+  command = "cd %s && git pull && %s _data/import-data && git add _data/team.yml _data/projects.yml && git commit -m 'update team and project data' && git push origin staging && %s go build >> %s" % (current, ruby, ruby, log)
```
konklone: Why have the staging site do the data extraction and git pushing? It seems like deploys are more predictable (though less automatic) when doing this work locally. Even updating the fabric script to run the git download and add commands on the local machine (while still doing everything else remotely) would be more preferable, I think.

gboone: I was uneasy about that from the beginning. I'm not a fan of non-people committing to our site. We can always have the deployment just update the data file but not commit it back and leave it to developers to update their data files as needed.

go
```diff
@@ -65,7 +65,7 @@ def update_gems
 end
 
 def update_data
-  exec_cmd 'cd _data && ./import-public.rb'
+  exec_cmd 'cd _data && ruby import-public.rb'
```
mbland: This could just be `_data/import-public.rb`. The explicit ruby interpreter is only necessary in the staging/prod environment.

deploy/fabfile.py
mbland: We do this for the internal Hub, but it's a much different use case. Everything automatically merged into the internal Hub, however, is still manually merged into the public Hub.

mbland: Also, @gboone, remember how I once added this to the `./go` deployment command:

```shell
if git commit -m 'update team and project data'; then git push origin staging; fi
```

Otherwise `git commit` will fail and `go build` won't run.

gboone:
Okay, I think this should be good to go now.

deploy/fabfile.py
```diff
@@ -33,26 +33,43 @@
...
 
 def restart():
-  run(
-    "cd %s && forever restart deploy/hookshot.js -p %i -b %s -c \"%s\""
-    % (current, port, environment, command)
-  )
+  with cd(current):
+    run(
+      "forever restart deploy/hookshot.js -p %i -b %s -c \"%s\""
+      % (current, port, environment, command)
```
mbland: I think `current` doesn't belong in this string interpolation tuple.

deploy/fabfile.py
```diff
@@ -33,26 +33,43 @@
...
+    )
+
+def update_data():
```
mbland: Ohh... I see what you did here. Nice!

go
```diff
@@ -64,8 +64,12 @@ def update_gems
   exec_cmd 'bundle'
 end
 
+def update_submods
+  edec_cmd "git submodule update --remote"
```
mbland: s/edec_cmd/exec_cmd/

deploy/fabfile.py
```diff
@@ -33,26 +33,43 @@
...
+    run("%s go update_submods" % (ruby))
+    run("%s go update_data" % (ruby))
+
+def update_submods():
+  with cd(current):
+    run("%s go update_submods" % (ruby))
+
+def build():
+  with cd(current):
+    run("%s go build" % (ruby))
```
mbland: Should this be `server_build` instead?

go
```diff
@@ -85,6 +89,15 @@ def ci_build
   puts 'Done!'
 end
 
+def server_build
```
mbland: What's the use case for `server_build`? Should it be used in `fabfile.py`? Also, both `server_build` and `update_submods` should have entries in the `COMMANDS` array. (You can steal the `CommandGroup` class from the Hub's `go` if you want. Or maybe it's time I wrote a `go` gem. :-P)

go
```diff
@@ -64,8 +64,12 @@ def update_gems
   exec_cmd 'bundle'
 end
 
+def update_submods
+  edec_cmd "git submodule update --remote"
+end
+
 def update_data
-  exec_cmd 'cd _data && ./import-public.rb'
+  exec_cmd '_data/import-public.rb'
```
mbland: For this to work on the remote server, I think you're gonna have to pass the path to the ruby interpreter.

go
gboone: Yeah, I think I meant to put this in the fabfile.

go
gboone: Even if it's called `/full/path/to/ruby go update_data`?

mbland: Yep. `exec_cmd` is a subshell.

deploy/fabfile.py
gboone: no, but it might be worth having a parameter to trigger `go server_build` like `build('full')` does the the full pull, update submodules, build the site but `build()` only rebuilds what's already there.

go
gboone: at 71b2f36 I did it this way, do you think it will work?
```
def update_data
  ruby = exec_cmd 'which ruby'
  exec_cmd "#{ruby} import-public.rb"
end
```
Trying to write it in such a way that if `ruby` ever moves the script will still work.

deploy/fabfile.py
```diff
@@ -33,26 +33,46 @@
...
+def update_data():
+  with cd(current):
+    run("%s go update_submods" % (ruby))
+    run("%s go update_data" % (ruby))
+
+def update_submods():
+  with cd(current):
+    run("%s go update_submods" % (ruby))
+
+def build(scope = 'simple'):
```
mbland: What's going on with `scope ='simple'`? And where is `full` defined?

gboone: updated see lines 75 and 77 below

gboone:
Hooray, Travis is passing again!

deploy/fabfile.py
```diff
@@ -33,26 +33,46 @@
...
+  with cd(current):
+    if scope = 'full'
```
mbland: Python needs `:` at the end of `if` statements.

gboone: Been `ruby`in' too much, I guess.

go
gboone: It's now used in the fabfile ;)

deploy/fabfile.py
```diff
@@ -33,26 +33,46 @@
...
+      run("%s go server_build" % (ruby))
+    elif scope = 'simple'
```
mbland: Not a blocker, but consider this instead:

```python
def build(scope = ''):
  with cd(current):
    if scope == 'full':
      run("%s go server_build" % ruby)
    else:
      run('%s go build' % ruby)
```

Not only does eliminate the extra conditional (which should've been `==`, btw), we eliminate the dreaded `s`-word. :-)

script/post
```diff
@@ -23,6 +23,20 @@ else
 	date = Date.parse(options[:date]).strftime("%Y-%m-%d")
 end
 
+def readContent(file)
```
mbland: Mind adding a comment explaining this little guy?

gboone: see: 1c0d6b9

script/post
```diff
@@ -23,6 +23,35 @@ else
...
+# append that path to the current file.
+#
+# The script will return an error to STDERR if the passed parameter is not a 
+# markdown file or if the file does not exist. Otherwise it returns the 
+# contents of the passed file.
+#
+# Author:: Greg Boone (mailto:gregory.boone@gsa.gov)
+# Copyright:: To the extent possible the contents of this method and the entire 
+# 			 script are public domain works
+# License:: CC0
```
mbland: Whoa, now _that's_ a comment. :-P But where is it used?

go
```diff
@@ -56,7 +56,7 @@ def init
     exec_cmd 'gem install bundler'
     puts "Bundler installed; installing gems"
   end
-  update_gems
+  update_gems
```
mbland: Nit: Trailing whitespace.

go
```diff
@@ -64,8 +64,13 @@ def update_gems
   exec_cmd 'bundle'
 end
 
+def update_submods
+  exec_cmd "git submodule update --remote"
+end
+
 def update_data
-  exec_cmd 'cd _data && ./import-public.rb'
+  ruby = exec_cmd 'which ruby'
```
mbland: Heh, neat trick. I'll consider doing that in the Hub's go script to avoid hardcoding the path.

script/post
gboone: See line 72, 1c0d6b9

go
gboone: We'll see if it works.

PR #515, opened 2015-02-09
Add byline to HSTS post, fix formatting in lists

konklone:
This fixes `<code>` formatting inside `<li>` tags, and adds a byline to the HSTS post.


gboone:
I think you have `base` and `HEAD` swapped.

konklone:
o_o

PR #516, opened 2015-02-09
Fix formatting and add byline to HSTS post

konklone:
This fixes `<code>` formatting inside `<li>` tags, and adds a byline to the HSTS post.


gboone:
LGTM

PR #517, opened 2015-02-09
Optimize all blog images

waldoj:
Basically all of the images were much larger than required, so I used ImageOptim to losslessly optimize them (that is, I didn't alter anything visible about the images).

Note the presence of commit 2815c65. I have no idea what internal bit of Git got coughed up there, or what to do with it, but good luck there.


gboone:
Thanks @waldoj sorry it took us a couple weeks to merge this.

PR #522, opened 2015-02-12
Fix the Hub API's content-type for JSON

mbland:
This is in reference to both 18F/hub#99 and 18F/hub#145. I also updated `/etc/nginx/nginx.conf` to update `gzip_types` to include `application/json application/javascript`. The changes are already running in production, and confirmed via [Postman](https://chrome.google.com/webstore/detail/postman-rest-client/fdmmgilgnpjigdojojpjoooidkmcomcm?hl=en).

cc: @gboone @konklone

### DIFF
--- a/deploy/18f-site.conf
+++ b/deploy/18f-site.conf
@@ -43,6 +43,7 @@ server {
 
   location /hub {
       root /home/site/production/hub/_site_public;
+      index index.html api.json;
       default_type text/html;
   }
 


gboone:
:+1: Thanks for catching this, @mbland and @janearc

konklone:
To be clear though, this should really go into the server's `/etc/nginx/nginx.conf` (which isn't in version control) so that it affects all endpoints.

PR #528, opened 2015-02-13
Blog instructions

gboone:
Greatly simplified blogging instructions per the new workflow we introduced a couple weeks ago. @18F/rearchitects your feedback is welcomed.

Also closes #511

PR #536, opened 2015-02-18
Deploys 535, 533, and 528 to production

gboone:
See #535, #533 and #528, chiefly, this deploy's the interview Steve Portigal post to production.


gboone:
Review here and merge when we're satisfied with the post on Staging

PR #552, opened 2015-02-25
Use JSON data from the public Hub API

mbland:
Replaces the need for the `_data/private` git submodule. When 18F/hub#182 is pushed to the public Hub, I'll follow through and add the imported `team.json` and `projects.json` files.

cc: @gboone


mbland:
The Travis failure would be because the JSON files aren't present yet. :-)

mbland:
With public Hub push 18F/hub#189 now live, I've imported `_data/projects.json` and `_data/team.json` and added them to this here PR. Looks great on my machine!

gboone:
Looks good.

PR #555, opened 2015-02-26
clarify that we aren't limited to the listed locations

afeld:
Just in case that would cause any confusion!

/cc @kaitlin @shawnbot @quepol @kategarklavs


gboone:
Where did we land on this?

elainekamlley:
@gboone @melodykramer Could this be closed since we have been speaking more about our distributed teams in blog posts? Or should we keep this open?

quepol:
I believe so, yes.

PR #557, opened 2015-02-26
Several pull requests worth of production deploys

gboone:
What it says on the can.


gboone:
cc: @kategarklavs if it looks good on staging, feel free to hit the green button

PR #572, opened 2015-03-02
Fix typos and clarify a few things.

monfresh:
This is for you, @melodykramer  and @gboone.

You'll notice that some changes are just cleaning up trailing whitespace. That's because my Sublime Text is configured to automatically delete them when I save a file, using the [TrailingSpaces](https://github.com/SublimeText/TrailingSpaces) plugin, which I highly recommend.

I also wanted to ask about this statement: `If you follow these instructions for any project you're working on in GitHub, they will work the same way`. Did you mean that the reader should expect pull requests to fail for other projects, or to work for other GitHub projects?

If the latter, that would not be entirely accurate because only people who have write access to a repo can submit pull requests after cloning a repo. No one can submit a PR to any repo they don't have write access to if they just clone the repo on their computer.

In order to be able to send a PR to a repo you don't have rights to, you need to *fork* it to your own GitHub account, then clone *your own fork*. All changes will be pushed to your own fork, and when you visit your fork, GitHub will let you choose whether you want to submit this pull request to the original repo you forked or your own fork.

### DIFF
--- a/_posts/2015-02-24-how-to-use-github-and-the-terminal-a-guide.md
+++ b/_posts/2015-02-24-how-to-use-github-and-the-terminal-a-guide.md
@@ -1,7 +1,7 @@
 ---
 title: "How To Use GitHub and the Terminal: A Guide"
 layout: post
-author: 
+author:
 - melody
 - boone
 tags:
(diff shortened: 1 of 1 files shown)


gboone:
@monfresh we're planning to add a disclaimer at the top of the post telling people that certain parts of the guide only apply to 18F team members. This is close to but not quite a finished post.

PR #578, opened 2015-03-04
Gah! We missed a step!

gboone:
Thank goodness for crowdsourcing! We caught this missing bit after [a conversation on Twitter.](https://twitter.com/heatherfro/status/572899679505133571)


gboone:
Other things to fix:
- GitHub has no app for Linux, though there are Linux GUIs for Git
- There are alternatives to GitHub, we should spell them out and maybe offer some examples of how to work with other remotes

PR #579, opened 2015-03-04
Some typos and link tweaks

konklone:
A few typos, and link tweaks.

I also clarified that "we" don't use Yosemite (18F does not manage this process, at least not yet), but that the post does.


gboone:
@konklone thanks or this, at the risk of merge conflicts I'm going to incorporate your changes here into #578 and close this out. Cool?

konklone:
Do with my commit what you think best.

gboone:
Okay, this all was incorporated with updates we pushed yesterday. Thanks @konklone

PR #585, opened 2015-03-04
Move Discovery fixes to production

konklone:
This moves the triple-copy resolution PR to production.


gboone:
Thanks for the quick action @elainekamlley and @konklone

PR #586, opened 2015-03-05
Deploys 575, 563, and optimizes images on production

gboone:
Chiefly, publishes: _Making Procurement Easier: Some Questions For Developer Kaitlin Devine_


gboone:
cc: @18F/blog

PR #601, opened 2015-03-12
Fix O*NET images and formatting

konklone:
The O*NET blog post's images are broken, and have incorrect formatting. This fixes all of those. Did this get eyeballed on staging before deploy?


gboone:
Yeah, it was on staging all night. Somehow we missed those photos. Thanks for the fix, @konklone.

konklone:
:+1:

PR #609, opened 2015-03-13
Deploy how to protosketch post

gboone:
Await @18F/blog, and author: @RobertLRead and @adelevie approval


gboone:
We may get an image, but I want to ship this before it gets too late at the end of the week.

PR #622, opened 2015-03-18
Sunshine Week: Extractive Industries Transparency Initiative Event

gboone:


gboone:
cc: @18f/blog @meiqimichelle the event starts at noon in DC, be good to get this out before then.

PR #627, opened 2015-03-18
Replace manual data download script with automatic plugin

dhcole:
This replaces the manual script and local `_data` files with a plugin that downloads each of them on build. Configuration for the plugin is set in the `_config.yml` file:

```yml
# get data from the hub
jekyll_get:
  - data: team
    json: 'https://18f.gsa.gov/hub/api/team/'
  - data: projects
    json: 'https://18f.gsa.gov/hub/api/projects/'
  - data: pif_team
    json: 'https://18f.gsa.gov/hub/api/pif_team/'
  - data: pif_projects
    json: 'https://18f.gsa.gov/hub/api/pif_projects/'
```

@juliaelman This will give you access to all of the team data in the hub, updated each time the site is rebuilt. You can access this data as if it were a local data file, with `site.data.team` in your templates.

This method should give you all the data you need for the new team layouts.

I know there's an open question about whether team bios should be the same on this site as the hub. With this, you have access to each member's hub bio. If you want to override that with something different, I suggest using the team member markdown files you started and using those instead of the bios here. 

Refs 18F/hub#220

cc @gboone @meiqimichelle @mbland


gboone:
Yes please thank you.

_plugins/jekyll_get.rb
```diff
@@ -0,0 +1,16 @@
...
+  class Generator < Jekyll::Generator
+    def generate(site)
+      config = site.config['jekyll_get']
+      if !config
+        return
+      end
+      if !config.kind_of?(Array)
+        config = [config]
+      end
+      config.each { |d| site.data[d['data']] = JSON.load(open(d['json'])) }
```
mbland: Here's the version I've drafted for the Hub, to ensure that fetched JSON is merged with whatever's already in `_data`, and that this generator gets run before anything else:
```ruby
require 'hash-joiner'
require 'open-uri'

module Jekyll_Get
  class Generator < Jekyll::Generator
    safe true
    priority :highest

    def generate(site)
      config = site.config['jekyll_get']
      if !config
        return
      end
      if !config.kind_of?(Array)
        config = [config]
      end
      config.each do |d|
        target = site.data[d['data']]
        source = JSON.load(open(d['json']))
        if target
          HashJoiner.deep_merge target, source
        else
          site.data[d['data']] = source
        end
      end
    end
  end
end
```

dhcole:
@mbland nice. Do you want to push an update to use your version of the plugin? Also, do we need any other updates to the build scripts to instead the `hash-joiner` gem?

mbland:
@dhcole Just pushed a commit. `hash-joiner` is already in the Gemfile. Give it a spin?

konklone:
Does making HTTP requests during build mean that working on the site while offline is not feasible?

gboone:
Good Q, @konklone I think you'd still be able to work on it just maybe with stale data as long as the plugin only rewrites the files upon a successful HTTP request.

konklone:
Sounds like it might hang for a while before timing out on the HTTP request in an offline environment? Don't mean to be a downer - this sort of situation is just the kind of tradeoff that static site generators make.

gboone:
Yeah, and I think I was wrong. @dhcole can you clarify: We're not actually downloading any files, just jamming the data into the `site` object at build time? If that's the case, what happens if the external API goes away?

dhcole:
With Mike's update, this plugin merges the data with whatever's in the site data hash. So if you want to work offline, just have a local copy of the data in your _data folder (the way it works now).

Yes, the plugin fails as currently written without network access but you could easily disable while working locally. Or, if you know of some better way to do this gracefully (something like a ruby version of a try...catch block) I can add that to the plugin.

dhcole:
Updated to optionally cache remote data sources in the `_data` directory. Now works from underground bunkers.

dhcole:
Updated `_config.yml`

```yml
# get data from the hub
jekyll_get:
  - data: team
    json: 'https://18f.gsa.gov/hub/api/team/'
    cache: true
  - data: projects
    json: 'https://18f.gsa.gov/hub/api/projects/'
    cache: true
  - data: pif_team
    json: 'https://18f.gsa.gov/hub/api/pif_team/'
    cache: true
  - data: pif_projects
    json: 'https://18f.gsa.gov/hub/api/pif_projects/'
    cache: true
```

_plugins/jekyll_get.rb
```diff
@@ -0,0 +1,39 @@
...
+          else
+            site.data[d['data']] = source
+          end
+          if d['cache']
+            path = "#{site.config['data_source']}/#{d['data']}.json"
+            open(path, 'wb') do |file|
+              file << JSON.generate(site.data[d['data']])
+            end
+          end
+        rescue
```
mbland: It'd be nice to emit an error message to note that a fetch/write failed, and from which endpoint/to which file. In fact, I think it'd be worth re-raising and allowing the build to fail.

_plugins/jekyll_get.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+      if !config.kind_of?(Array)
+        config = [config]
+      end
+      config.each do |d|
+        begin
+          target = site.data[d['data']]
+          source = JSON.load(open(d['json']))
+          if target
+            HashJoiner.deep_merge target, source
```
mbland: So there is one bit of danger here. If `target` contains any non-empty Array members that are duplicated in `source`, those duplicates will appear in `target`. Since the original intent of this plugin was merging data from Midas with data stored in the Hub, this makes sense. But keeping a full cached copy of the data in the repo is a different use case.

Hence, I'm wondering if the cached version of the data should live somewhere like `_data/cached`. Then, in this plugin, in the `!config` case, we can use: `HashJoiner.promote_data site.data, 'cached'`.

_plugins/jekyll_get.rb
```diff
@@ -0,0 +1,39 @@
...
+require 'open-uri'
+
+module Jekyll_Get
+  class Generator < Jekyll::Generator
+    safe true
+    priority :highest
+
+    def generate(site)
+      config = site.config['jekyll_get']
+      if !config
```
mbland: What do you think about: `if !config || config['disabled']`?

_plugins/jekyll_get.rb
dhcole: `_data/cached` is non-standard though. I'd rather not introduce new conventions. We can keep this simple. The data that this would cache accounts for merging remote data with local data (but only if you choose to cache).

Since caching is optional to begin with, I think this behavior is fine.

_plugins/jekyll_get.rb
dhcole: Just as a way to disable without deleting the config? We could do that, but let's not hold this up any further for discussions about the plugin. The goal was to validate that it solves our problem (getting data from the hub on build). Since that seems answered, we should get this plugin into a separate repo where we can improve it with additional ideas like this.

_plugins/jekyll_get.rb
mbland: `d['cache']` could be the path to the cached version of the data. I don't see anything wrong with adding new conventions if they solve real problems.

mbland:
OK, let's merge this now, but let's also get that plugin-specific repo set up ASAP.

dhcole:
:+1: thanks @mbland 

Conversation copied over to here: https://github.com/18F/jekyll-get/issues

Issue #430, opened 2014-12-18
create a tag index page? 


gbinal:
https://18f.gsa.gov/tags/ doesn't resolve but was where I went to look to try and find a listing of what tags there are.

gboone:
Bumping this. Would like to see not only a list of tags but context around what they are (example posts? descriptions in a `_data` file? Maybe don't include all tags but most used?

gboone:
inspiration: http://alistapart.com/topics

melodykramer:
love the List Apart example.

elainekamlley:
Closes with #875, https://18f.gsa.gov/tags/

PR #654, opened 2015-03-27
replaces tags plugin with jekyll-archives

### LINKED ISSUE #653: Swap custom tag archive generator for jekyll-archives
jekyll-archives already supports the 3.0.0 beta and gives us a lot of the back end work we're looking for in #155 and #182

gboone:
Replaces the custom-built tag archive generator with the Jekyll/jekyll-archives plugin. Closes #653 and makes #155 and #182 more possible. Also allows us to quickly add date and category archives in the future should we ever need that.

In addition, this PR also standardizes some of the tags on capitalization because they were not generating correctly. The How We Work tag, for example, was listed in some posts capitalized and in others lower case. This inconsistency caused the archive to only surface one of the posts instead of all 28. It's possible this bug also existed in our hand-crafted tag archive generator.

There are a few upshots to switching to this plugin and retiring ours.
1. Tag archives will have better URLs because they will no longer have spaces in them.
2. We get the archive landing page out of the box making it easier to accomplish #430.
3. The plugin gives the archive page object a `posts` entry so we can get rid of a lot of unnecessary complexity from the template.
4. It's an open source project and comes with the advantage of a larger number of people working on it than our plugin ever will.
5. There's probably a fifth I haven't thought of yet.

**Note:** this introduces a new gem and will require users `bundle install` to get the new hotness


gboone:
This will  also close #430

PR #655, opened 2015-03-30
Optimized all the images

gboone:
I think this will reduce page size on some of our pages.


konklone:
I just want @waldoj to know that one person truly can make a difference.

waldoj:
I guess I would say I'm like Gandhi, only instead of making my own salt, I'm dropping some images onto ImageOptim, and instead of weaving my own fabric, I'm making GitHub commits, and instead of threatening my own wellbeing in any way, I'm making snarky comments.

"A convert's enthusiasm for his new compression algorithm is greater than that of a person who is born in it."

gboone:
Gotta say, having imageoptim fire off every time I run `git commit` sure is a good reminder.

PR #656, opened 2015-03-30
First pass at 18f-consulting's post for this morning

gboone:
Still working out some formatting details, @18F/blog want to give it a look over?


gboone:
Alright, I think we're good to go on this, let's proof it on Staging.

Issue #633, opened 2015-03-19
Automate bylines


gboone:
Right now blog posts require the following to get a simple byline:
```yaml
# in the frontmatter
authors:
- author1
- author2
```
and in the post body

```
<p class="authors">
  by {% author author1 %} and {% author author2 %}
</p>
```

Let's eliminate the redundancy and see if it's possible to auto-generate the byline in the post template. We accidentally publish too many posts without bylines.

Desired state:
Something like this in `layouts/post.html`

```
<p class="authors">{% authors %}</p>
```

Considerations: 
- some of our posts have illustrators
- authors are available in the site object at: `site.data.authors` in templates. There are different ways to tap into this depending on what kind of plugin you're making.
- there are probably other Jekyll sites that have done this; [I did it for a side project of mine](https://github.com/InternationalUnderground/internationalunderground.org/blob/master/_plugins/feature.rb#L143-L146), but that might not be the best model.

konklone:
Some of our posts have authors not listed in the team data file, and include their title. We could make it so that if an author is not found by the handle given in the front matter array, it is instead treated as plain text to be embedded, and not a lookup handle.

gboone:
Great idea, @konklone, thanks for the suggestion. :+1:

konklone:
Pretty sure this is done with #966, right?

PR #669, opened 2015-04-02
Should stop jekyll-get on local

### LINKED ISSUE #640: Option to disable jekyll-get in certain environments
Esp. on slower connections, `jekyll-get` fetching all the data it does can add several seconds to build time.

slower connection:
```
      Regenerating: 1 file(s) changed at 2015-03-20 10:26:39 ...done in 8.639929 seconds.
      Regenerating: 4 file(s) changed at 2015-03-20 10:26:48 ...done in 8.339296 seconds.
      Regenerating: 6 file(s) changed at 2015-03-20 10:26:56 ...done in 10.461097 seconds.
```
normal:
```
      Regenerating: 1 file(s) changed at 2015-03-18 10:50:16 ...done in 3.690993 seconds.
      Regenerating: 1 file(s) changed at 2015-03-18 10:52:11 ...done in 4.006066 seconds.
      Regenerating: 3 file(s) changed at 2015-03-18 11:35:38 ...done in 4.453949 seconds.
```

gboone:
Should close #640 by disabling the plugin in local dev environments. See that and 18F/jekyll-get#5 for more context on why we need this disabled on local environments.

This also restores the import-data script for folks who might want to refresh the data files locally. Since the plugin will still run in production environments, the data should always refresh with each site re-build. 

Data can be refreshed in local environments in either of two ways:
1. By adding `force_get: true` to `_config.yml`
2. Running `_data/import-data.rb`

Pardon the non-descriptive, misleading branch name, I started it attempting to attack #633 and ended up doing this as a result.


_plugins/jekyll_get.rb
```diff
@@ -15,6 +15,9 @@ def generate(site)
       if !config.kind_of?(Array)
         config = [config]
       end
+      if site.config['host'] == "127.0.0.1"
```
konklone: I typically access the site on `localhost`, not `127.0.0.1` -- or through an `ngrok` or `localtunnel` URL. Pardon my Jekyll ignorance, but how does `config['host']` get filled in?

gboone: Yeah, I'm not sure. I don't think ngrok would affect it unless you overrode it in `_config.yml` or with `--host` on the command line. Now that I think it through, I need to check to see if `host` gets set on `jekyll build` or this would probably stop it on staging/production as well.

gboone:
After conversing with @konklone decided to forego modifying the `jekyll-get` plugin in favor of a more flexible solution of using a second config file for deployment environments. This pull request now commits the json files, re-adds the `import-data` script (so that folks can refresh data if needed), and updates the `go` and deployment scripts to reflect the dual configuration files to be used in deployment environments.

Folks who want refreshed data have two options now:
- run the `import data` script
- serve the site with `./go server_build` followed by `./go serve`

_deploy_config.yml
```diff
@@ -0,0 +1,14 @@
+# get data from the hub
```
konklone: I suggest renaming the file to `_config-deploy.yml` or `_config_deploy.yml`, for consistency and alphabetization's sake.

gboone: good idea, thanks

konklone:
This is a much better approach, and probably the best approach to use `jekyll-get` during the build process.

However, I do think the work of `jekyll-get` is not best done during the build process. It adds delay to the deploy process, and makes production deploys dependent on successful external network connections. I think our deploys should be quick, deterministic processes. 

Additionally, this whole system is meant to prevent us from having to do manual work to sync data from our central data repository into this repository, but this doesn't fully solve that problem. When upstream data is changed, nothing automatically causes our site to re-deploy, and so the site will remain out of date until the next deploy is manually triggered during a content change.

Ideally, a fetch of remote data would only happen when changes to upstream data happen, and some sort of ping would get sent to fetch and rebuild. Or perhaps more ideally, when upstream data changes occur, it would sync those changes automatically to a copy of the data in this repository (similar to the file we used to have) which would cause a quick deploy using the existing webhook process.

Less ideally, we'd just refetch and rebuild the site at least once overnight or something, and that's the only time we'd turn refetching on. I think either that approach, or a downstream ping approach, would be better than building jekyll-get into the production deploy.

That all said! :+1: on this PR as improving the process and greatly easing local development.

gboone:
Alright, gonna merge this. Thanks for the consults, @konklone. I agree that build time might not be the best time to run this but for now it's a step forward until we have time to build something more robust.

PR #670, opened 2015-04-02
Cage match

melodykramer:

PR #673, opened 2015-04-03
Adds /consulting

gboone:


elainekamlley:
Awesome. The icons for twitter and GitHub are broken.

gboone:
Good catch, Elaine, I'll push an update to fix that.

gboone:
Okay, should be fixed after we push #670 this afternoon

PR #677, opened 2015-04-03
Actually reference the new fonts

konklone:
Ugh - I hadn't committed the change to the header file, so the `<link>` tag was still pointing to `fonts.googleapis.com`.

@gboone While my local testing was sound on Chrome and Firefox, this means that your tests on Safari and IE were not actually testing the inlined fonts. My apologies. Mind running that test again?

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -81,7 +81,7 @@ http://updates.html5rocks.com/2014/11/Support-for-theme-color-in-Chrome-39-for-A
 
 <!-- CSS
 ================================================== -->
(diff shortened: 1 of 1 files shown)


konklone:
Note I got into this situation because I was trying to ignore a bunch of other local changes involved in me temporarily disabling `jekyll-get` because of the large build times. So I'm looking forward to the repo defaulting to not doing that. :)

gboone:
@konklone I patched up #669 if you're interested

PR #679, opened 2015-04-03
Improving font performance.

msecret:
This method asynchronously loads fonts as soon as possible, while not
blocking render, then stores the result in localStorage for future
visits.

This has been documented here:
http://www.filamentgroup.com/lab/font-loading.html

TODO, there could be improvements made to this later down the line:
- Separate each type of font file (ttf, woff) into a separate file
with font data base64 encoded. Program if statement for different types
- Use font loading events to apply font class after font loaded.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -79,15 +79,57 @@ http://updates.html5rocks.com/2014/11/Support-for-theme-color-in-Chrome-39-for-A
 <!-- Mobile "theme" -->
 <meta name="theme-color" content="#1188ff" />
 
+<!-- Fonts
+================================================== -->
+<noscript>
+  <link href="/assets/css/google-fonts.css" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


konklone:
> * Use font loading events to apply font class after font loaded.

I don't see this evident in the code - can you point out where font events are used? (I do see an XHR event.) I also don't see any classes used here.

msecret:
Sorry for not being clear, the bottom two points are what could be done in the future for improved performance.

I didn't do them this time to keep the code compatible with your changes and also because I haven't tested the other method yet.

The solution I have now is good enough performance. I'm going to try and develop a slightly better method in the future.

konklone:
Ahhhh, I skimmed your original issue description too fast, I'm sorry. -_-

Looks good to me!

konklone:
Am I right in assuming this method of font optimization works whether or not the fonts have been inlined, or loaded dynamically from Google Fonts? Does having inlined them make any difference to the approach?

PR #688, opened 2015-04-09
Remove category from Peace Corps post

konklone:
The Peace Corps post URL is prefixed with `/design%20and%20code/`:

> https://staging.18f.us/design%20and%20code/2015/04/09/flexibility-when-releasing-a-new-product-peace-corps-new-donation-platform/

It should be: 

> https://staging.18f.us/2015/04/09/flexibility-when-releasing-a-new-product-peace-corps-new-donation-platform/

It's because of the `category` field in the front matter, which this PR removes. I think we need to fix the URL generation before we can start using categories.


gboone:
@melodykramer we'll have to pay attention to this while we're working through #654

melodykramer:
@konklone good catch.

PR #689, opened 2015-04-09
Use pa11y to test homepage

adelevie:
Ignore this for now. I'm testing out how we can have some automated accessibility testing for the site.


gboone:
would this make #566 obsolete?

adelevie:
I'll just say that #566 is obsolete. Blast it away!

gboone:
:boom:

konklone:
Curious the state of this, and of pa11y?

gboone:
I was about to say the same thing.

gboone:
We still want to do this; there was a good discussion in Slack about how we might do it, and how we might use the results effectively. We'd also have to re-open this against `master` so that we're testing the new site. Closing for now.

PR #691, opened 2015-04-10
Deploy RFP-EZ notes to production

konklone:
This deploys @aaronsnow's changes to production, currently visible at https://staging.18f.us/rfpez/.


gboone:
This will now also deploy emoji support and @jenniferthibault's photo, as of #717 and #695

PR #721, opened 2015-04-11
Fix font loading

konklone:
It got broken when converted to JS, it's supposed to include a `.min` in the file name.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -120,7 +120,7 @@ http://updates.html5rocks.com/2014/11/Support-for-theme-color-in-Chrome-39-for-A
 
     // Site fonts to load.
     loadFont('google', '/assets/css/google-fonts.css');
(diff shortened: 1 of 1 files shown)


konklone:
Hold off on merging this -- there's a subsequent bug that should also be fixed.

@msecret, it looks like the font loading method you moved font-awesome.css to breaks what the stylesheet considers its relative font resolution path. So `url('../fonts/...')` used to refer to `/assets/fonts`, relative to the CSS file itself, but now refers to `/fonts`, relative to the index page the JS runs on. Any ideas?

konklone:
I've updated this PR to revert #679 for the time being, and remove the JS-driven font loading optimization, as it broke the relative `url()` paths contained in `google-fonts.css` and `font-awesome.min.css`.

cc @msecret

PR #727, opened 2015-04-13
Deploy a few bug fixes to production

gboone:
Chiefly, this will fix the sorting problem on the homepage and the responsive layouts of the team members


konklone:
It looks like @jeanninehunter's image isn't 250x250?

PR #729, opened 2015-04-14
Fix name wrapping on bigger screens

shawnbot:
My fix from yesterday introduced a regression at bigger sizes. Widths between about 980 and 1100px exhibit this:

image

This tweak moves the `min-height` for the name elements out of the smaller breakpoint so that it's always applied.


juliaelman:
Looks good to me! :+1: Merge away

PR #730, opened 2015-04-14
Require rouge, remove Python dependency

konklone:
Rouge wasn't being properly referenced or required, so Pygments was still being used for syntax highlighting.

Confirmed that syntax highlighting works identically at:

> http://localhost:4000/2014/11/17/taking-control-of-our-website-with-jekyll-and-webhooks/


juliaelman:
Verified fixe! Thanks @konklone :rocket:

gboone:
Thanks @juliaelman for finding the bug and @konklone for fixing it!

PR #732, opened 2015-04-15
Handle 18f.gov redirect

konklone:
https://18f.gov -> https://18f.gsa.gov

### DIFF
--- a/deploy/18f-site.conf
+++ b/deploy/18f-site.conf
@@ -207,11 +207,16 @@ server {
     return 301 https://staging.18f.us$request_uri;
 }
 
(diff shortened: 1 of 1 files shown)


gboone:
These are already deployed; this is just an update FTR. Correct?

gboone:
These are already deployed; this is just an update FTR. Correct?

konklone:
Yep. I try to be rigorous about never editing 18f-site.conf on the server. I edit it locally, then copy it to the server and test/deploy it. Then I commit the changes locally, knowing that the server copy reflects it.

PR #735, opened 2015-04-17
Fixit day sprint

gboone:
- upgrades to Jekyll 3.0 (_finally_)
- adds gems appropriate for Jekyll 3.0 (which seems a little more verbose about missing gems, probably a good thing)
- updates data from the public version of the hub
- fixes humans.txt for Jekyll 3.0


konklone:
A note on Jekyll 3 -- I found bugs with `--watch` mode under Jekyll 3's new default of partial rebuilds.

gboone:
Oh, good catch. I think there's a way to force full rebuilds, if not will disable rebuilding by default.

gboone:
Okay, updated to force a full rebuild by default. If we have trouble with that we can completely disable it until the next version of Jekyll (beta or no) comes out.

gboone:
@konklone this look good now? Feel free to merge if :+1:

konklone:
:+1:

PR #747, opened 2015-04-21
removed Rob Klopp; added Kaitlin Devine

aaronsnow:


gboone:
@kaitlin is involved in this conference, excellent!

PR #758, opened 2015-04-23
Fix byline for Max post

konklone:
The byline needs to be explicit.


gboone:
Oh, good catch, thanks @konklone

PR #764, opened 2015-04-25
Fix a bunch of links, a few third party images

konklone:
This does a few things:

* Inlines a couple of third party images.
* Moves a bunch of old `18fblog.tumblr.com` URLs to their migrated `18f.gsa.gov` URLs.
* Moves a bunch of common URLs from `http://` to `https://`


gboone:
Thanks @konklone it'd be interesting to have a CI scan that would alert us to links we have in older blog posts as `http` that have `https` available.

konklone:
Agreed. @benbalter uses html-proofer, but that'd be a new mode.

PR #774, opened 2015-04-27
Fix the RSS Feed

gboone:
This should fix 18f.gsa.gov/feed and close #756 by reverting Jekyll to the most stable version.

PR #776, opened 2015-04-27
Deploy many light fixes, fix the RSS feed, and update a blog post

gboone:
This will send #764 #763, #774 and #775 to production, don't merge until all those have been proofed on staging.


gboone:
Confirmed the feed is repaired on staging. We can merge any ol' time.

PR #780, opened 2015-04-28
Publish: The Intersection of Art and Technology

gboone:
by @juliaelman 

For when we're ready on staging.


gboone:
This is ready and now includes many new team member photos. With those photos we've introduced a small css bug that seems to only affect Firefox users (or maybe just me ¯\_(ツ)_/¯), if someone wants to dig in and try fixing that this morning I will love you forever. :heart_eyes_cat: :heart_eyes_cat: :heart_eyes_cat:

gboone:
:mouse: :boom:

Issue #786, opened 2015-04-29
Strange CSS Behavior on not-Chrome browsers


gboone:
Introduced likely with #782 but perhaps it was always lurking just waiting for the right combination of photo + text to appear. Something seems to be causing Eric R's `.bio` element to push too far down the row, leaving a startling gap before Marco's `.bio` element. I suspect a `max-height` will fix it, but we'd have to make sure that doesn't introduce new problems on smaller screens of browsers that _don't_ have this problem.

Which browsers: all but Chrome and Safari. (Only tested on IE 11, assume it's also on earlier versions)
**Firefox:**
![Image of the affected row in Firefox](https://cloud.githubusercontent.com/assets/312468/7392902/cde2fb58-ee58-11e4-8c3b-eae885ab82c6.png)
**IE 11:**
![Image of the affected row in IE 11](https://cloud.githubusercontent.com/assets/312468/7392845/7fa67e2e-ee58-11e4-8c70-cd84d442f486.png)

PR #787, opened 2015-04-29
Fix grid bug

maya:
I resized Eric's image to 250px. This resolves https://github.com/18F/18f.gsa.gov/issues/786

This is how it looks in FF now...
screen shot 2015-04-29 at 8 54 26 am


gboone:
Thanks @maya and @elainekamlley

PR #790, opened 2015-04-29
Use Jekyll::Utils.deep_merge_hashes instead of HashMerger

parkr:
https://github.com/jekyll/jekyll/blob/master/lib/jekyll/utils.rb#L20

Also, rename `d` to something a bit more descriptive.

_Submitted via github.com, wasn't able to test this_

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,6 @@ group :jekyll_plugins do
   gem 'jekyll-paginate'
 end
 
-gem 'hash-joiner'
 gem 'safe_yaml'
 
 gem 'jemoji'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -25,8 +25,6 @@ GEM
     fast-stemmer (1.0.2)
     ffi (1.9.8)
     gemoji (2.1.0)
-    hash-joiner (0.0.7)
-      safe_yaml
     hitimes (1.2.2)
     html-pipeline (1.11.0)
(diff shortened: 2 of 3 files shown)


parkr:
Side note: Having to install Ruby 2.2.x just to build the site is frustrating, knowing that Ruby 2.2 didn't really diverge from 2.1 all that much. The error message could use some more details:

```shell
~/code/18f.gsa.gov#parkr-patch-1$ ./go

*** ABORTING: Unsupported Ruby version ***

Ruby version 2.2.0 or greater is required to build 18f.gsa.gov, but
this Ruby is version 2.1.6. Consider using a version manager such as
rbenv (https://github.com/sstephenson/rbenv) or rvm (https://rvm.io/)
to install a Ruby version specifically for 18f.gsa.gov development.
```

gboone:
@parkr I'll take a look at this, and thanks for the reminder that we don't need `jekyll_get` any more. (See: #725) I think we're using `hash-joiner` in other scripts (see _data/import-public.rb) so while we might be able to build the site without it, it might take some refactoring over there too.

Thanks for the note on forcing 2.2.x to build the site. It's there to make sure our developers are in the same environment. Perhaps we can make that print a warning instead of an exit.

gboone:
Also, you might suggest those same changes to [18F/jekyll-get](https://github.com/18f/jekyll-get)

Issue #760, opened 2015-04-24
Move staging site


konklone:
From https://staging.18f.us to https://staging.18f.gov.

PR #794, opened 2015-04-30
Add CF deployment scripts and manifest

dlapiduz:
These are some sample CF manifests and scripts that we can use to deploy the site.

I deployed it to `site.18f.gov` but you all can do what you see fit with this.

There is an org `18f` and a `site` space if you are not invited please ask me.

### DIFF
--- a/.cfignore
+++ b/.cfignore
@@ -1,0 +1,4 @@
+.git
+assets
+!_site/assets
+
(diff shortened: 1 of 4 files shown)


konklone:
Aside - I think we'd need `staging.18f.gov` to make it obvious that this is not "the 18F site" to anyone who accidentally ends up here.

gboone:
This does not close, but does get us closer to closing #760 we definitely need staging.18f.gov soon.

PR #801, opened 2015-05-04
adds author metadata to News section

alishalisha:
### What and Why
When I read the "News" section of 18F, I would love to quickly see who has written the post.

This PR looks for the author or authors, and lists them as secondary metadata as part of the blog post metadata block. If no author exists, no author block is printed.

### Screenshot
screenshot 2015-05-04 17 34 34

### Discussion
I noticed that you all have stored Author data in frontmatter, which led to some inconsistencies in spellings, capitalization, first name only vs. first-and-last-names listed, etc. I'm sure there is a smarter way to solve this problem if the author info is stored elsewhere, and would love to see what you may come up with. I don't know Ruby so I did what I could :smiley:

### DIFF
--- a/assets/_sass/_custom.scss
+++ b/assets/_sass/_custom.scss
@@ -388,6 +388,9 @@ $tiniest: new-breakpoint(max-width 390px);
 		margin-bottom: 0.6em;
 		line-height: 1.2em;
 	}
+	.quiet {
+		color: $medium-gray;
+	}
 }
 .tag-index {
 	.blog-title h1 {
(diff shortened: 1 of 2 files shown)


gboone:
This looks pretty good, @alishalisha. The names we list in the frontmatter (and what you're seeing in the news section) is a slug that corresponds to authorship information we keep in a data file (one that's machine generated) found at `_data/team.json`. On individual blog posts we use a plugin `{% author slug %}` to generate the full name but we're hoping to build a way to make that a bit smarter, see #633 for some more context around this.

PR #805, opened 2015-05-06
Adjust inlined web fonts to use WOFF2 where possible

konklone:
This follows the conversation with @mathiasbynens that began in https://github.com/18F/18f.gsa.gov/pull/672#issuecomment-97320631, and continued at https://github.com/mmastrac/webfont-dl/issues/8, and switches our inlined fonts to use WOFF2, a [recent improved font file format](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/j27Ou4RtvQI/discussion). 

I deleted the old downloaded font files, and followed @mathiasbynens' suggested command:

```bash
webfont-dl 'https://fonts.googleapis.com/css?family=Raleway:400,700%7COpen+Sans:400,700' -o assets/css/google-fonts.css --font-out=assets/fonts/ --css-rel=../fonts --woff1=link --woff2=link --svg=omit
```

This downloaded woff2 versions of the fonts, and changed the strategy to link to all of them, rather than inline them. This will load the correct font in all browsers. However, compared to the current approach, it will penalize some older browsers which do not support SPDY (most IE) by making them download additional files, and will not penalize [SPDY-supporting browsers](http://caniuse.com/#search=spdy). I like the tradeoff, given SPDY's wide support, and given that [browsers which support SPDY and WOFF2](http://caniuse.com/#search=woff2) (recent Chrome, Firefox, and Opera) will get improved speed at loading fonts and getting ready to go.

@gboone If you think this is a good direction, would you mind giving this a quick test in Internet Explorer?


mathiasbynens:
:+1:

gboone:
@elainekamlley this looks good to me, mergin'

Thanks @konklone and @mathiasbynens

elainekamlley:
:+1:

PR #817, opened 2015-05-11
Update README.md

wslack:
This pull request brings README into sync with the new process.


elainekamlley:
Hey @wslack, thanks for jumping in! I think we want to keep picture instructions, since we will want new hires to add their picture this way once the redesign is finished. But I definitely want to add the "answer bio questions" step. I'll take another pass on this.

elainekamlley:
After review, I think the site team needs to have review photos team members submits and make edits. Thanks @wslack for updating this.

PR #827, opened 2015-05-12
Fix RSS feed

konklone:
We broke the [RSS feed](https://18f.gsa.gov/feed/), it's now wrapping the contents in a default layout.

If I remove the recently added default values in `_config.yml`, it reverts to not having a layout. After a few different attempted workarounds, I added a `nothing.html` layout file, and set `layout: nothing` at the top of the `feed.xml` file.


quepol:
Merging on staging to test fix.

gboone:
Defaults!!! Thanks @konklone

PR #822, opened 2015-05-11
More deploy script fixes, edits to Micah's name

gboone:
Deploys #821 and #820 to production

PR #829, opened 2015-05-12
Fixes deployment script and RSS feed

### LINKED ISSUE #828: Mime type for RSS feed is HTML
I feel like it used to be a proper XML mime type? But even though the RSS feed is back to normal content [on staging](https://staging.18f.us/feed/), the mime type is `text/html`. This might inhibit compatibility with some clients. I don't know what to do about it while keeping the `/feed/` URL, except to add an exception rule in the nginx vhost.

gboone:
The RSS feed is still broken because it is not getting the correct MIME type. The `nothing` layout used to override the `defaults` set in `_config.yml` for posts. The "why" for defaults for posts affecting pages sounds like a Jekyll bug or we're doing it wrong. Could be either. This fixes it by setting `layout: false` on `pages/feed.xml`. 

Closes #828 #827 fixes bugs introduced in #822.

Also fixes additional problems with the deployment script, namely, when I created `_config-deploy.yml` I apparently forgot to put anything in there. I discovered this when attempting to do a full site rebuild and found unexpected errors with the generated pages.

cc: @konklone @quepol #learning


gboone:
This is somewhat urgent so if we could roll this before any blog posts get published that'd be great. @18F/blog @18F/rearchitects

konklone:
A note that you can delete the `nothing.html` layout now any time.

PR #830, opened 2015-05-12
RSS feed fix

gboone:
Still getting the wrong MIME type but it looks like browsers are at least picking up on it as an RSS feed instead of an HTML document.


gboone:
If we hold this until #831 we can get a full fix.

gboone:
This PR will now also publish the new blog post "Announcing the CALC Tool: Making Pricing Research Easier in Federal Procurement" (#834)

PR #831, opened 2015-05-12
Feed types

gboone:
@konklone I suggest pulling this branch down and running `curl -I localhost:4000/feed/` against this branch before we merge to double check my work. And maybe copying and pasting the resulting `_site/feed/index.xml` file into an RSS validator. 

When I ran it through a validator, we had one parse error and some areas for improvement.

When I `curl -I` the endpoint I got:

```
HTTP/1.1 200 OK 
Etag: bbdee8-23f06-55521294
Content-Type: text/xml
Content-Length: 147206
Last-Modified: Tue, 12 May 2015 14:47:48 GMT
Server: WEBrick/1.3.1 (Ruby/2.2.0/2014-12-25)
Date: Tue, 12 May 2015 14:56:11 GMT
Connection: Keep-Alive
```

And here I thought the RS in RSS was for "really simple." :smile_cat:


gboone:
I have not checked our nginx configuration to make sure it'll pick up on `index.xml` as an index file.

konklone:
:+1: Works for me locally.

Issue #815, opened 2015-05-11
When clicking "work" on the new homepage, I lose access to the menu.


wslack:
Jumping to the dashboard is great, but the loss of the menu (and without a clear "back to homepage button") disoriented me. Is it possible to add the menu to the dashboard page?

gboone:
It's possible.

awfrancisco:
Closing since there's no "work" link in current header.

PR #837, opened 2015-05-12
Add recruiting address to webpage footer (?)

wslack:
Adds a line for "join18F@gsa.gov" at bottom of webpage with "Hiring" in link. Not entirely sold this is needed, but if wanted or desired...here's the edit. I can buy the argument that its a bit confusing, and we do get messages to the general address regularly.

Test Edit for GitHub class: this should link to issue 815: https://github.com/18F/18f.gsa.gov/issues/815

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -7,6 +7,8 @@
       <h1>Contact Us</h1>
       <a href="mailto:18f@gsa.gov" itemprop="email"><i class="fa fa-envelope"></i> 18F@gsa.gov</a>
       <br/>
+      <a href="mailto:join18f@gsa.gov" itemprop="email"><i class="fa fa-envelope"></i> Hiring: join18F@gsa.gov</a>
+      <br/>
       <a href="https://twitter.com/18F"><i class="fa fa-twitter"></i> @18F</a>
     </div>
     <div>


gboone:
I think the plan in the works is to link out to [the Join 18F site on pages.18f.gov](https://pages.18f.gov/joining-18f/) cc: @jamiealbrecht @elainekamlley

gboone:
We're linking to https://pages.18f.gov/joining-18f from the homepage now. I think we still plan to offer some contact information in the footer so we'll leave this open.

jamiealbrecht:
The current thought is to have folks officially apply through USDS with the join18f referral code. I'd like join18f@gsa.gov for contact Team Talent or something or other for all hiring related questions. But not sure if that's confusing either?

gboone:
If I'm following: The join18f email address is for general inquiries but not applying. That makes me think we _shouldn't_ put that email address here but maybe on a more comprehensive contact us page.

jamiealbrecht:
Right. That's the direction we want to go. Of course, if candidates still email there with applications, that's fine. They will get into the process. However, want to move to a world where everyone comes through one funnel. 

That said, I love the idea of repurposing Join18F for contact our Talent Team. Although, not sure if it has value? I am not sure folks would use it?

TLDR, we need to think it out. :)

PR #838, opened 2015-05-12
Deploy tests

gboone:
We can definitely do better than this, but this is a good first push for at least testing if a file isn't getting generated that should be or if it has the wrong extension. These tests would have caught all the problems we had with the feed.

This PR also includes the nginx configuration changes we did on staging/production this morning to properly server the rss feed.


gboone:
Okay, the test runs correctly now and should exit zero if one of the test conditions fails. (It failed yesterday when the RSS feed was still broken on this branch. You can see the test output in the Travis build from e1f9fff

PR #853, opened 2015-05-19
Optimize a *ton* of blog images

gboone:
Average savings of about 6% overall ( > 70% per file)


gboone:
This now includes all the other images in our site that were not optimized. ImageOptim showed 0.00% optimization on the emoji but they still showed up ¯\_(ツ)_/¯

PR #862, opened 2015-05-26
Restore /hub/deploy webhook

mbland:
Also, per 18F/hub#299, I updated `/etc/nginx/nginx.conf` to use `gzip_static on` after rebuilding Nginx at version 1.8.0 with `--with-http_gzip_static_module` enabled, in anticipation of using gzipped files produced by the impending [jekyll_pages_api_search gem](https://github.com/18F/jekyll_pages_api_search/).

BTW, I noticed that a `sudo service nginx restart` isn't enough to ensure the new build is loaded; the error logs show `unknown directive "gzip_static"` when issuing `sudo nginx -s reload` after a restart. You have to manually kill the master process, then run `sudo service nginx start`. After that, the `gzip_static` directive warnings go away.

Manually reissued the last webhook to `/hub/deploy` to ensure that things are working again as intended.

cc: @gboone @konklone

### DIFF
--- a/deploy/18f-site.conf
+++ b/deploy/18f-site.conf
@@ -97,6 +97,22 @@ server {
       proxy_read_timeout    30;
   }
 
+  location /hub/deploy {
+    proxy_pass http://localhost:4002/;
+    proxy_http_version 1.1;
+    proxy_redirect off;
+
+      proxy_set_header Host   $host;
+      proxy_set_header X-Real-IP $remote_addr;
+      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+      proxy_set_header X-Forwarded-Proto https;
+      proxy_max_temp_file_size 0;
+
+      proxy_connect_timeout 10;
+      proxy_send_timeout    30;
+      proxy_read_timeout    30;
+  }
+
   # Redirect /code/<project> to the corresponding GitHub page
   location /code {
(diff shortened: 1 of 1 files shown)


konklone:
> You have to manually kill the master process, then run `sudo service nginx start`.

I assume you could also run `sudo service nginx stop` instead of manually killing the master process.

konklone:
Did this webhook get removed or something?

mbland:
You would think `sudo service nginx stop` would do that, but no. Eventually the master process gets respawned or something, so the PID used to try to restart the service has expired, and the service script silently fails.

Yeah, the webhook got removed somehow, probably because I don't think I ever checked the config in in the first place.

gboone:
The stopped webhooks might be due to something I've noticed recently. Running the `stop` command in the fabfile stops _all the webhooks,_ but `start` only starts the one you specify. It's possible when we were working through the deployment recently we stopped it on either dashboard, hub, or 18f-site and forgot to restart the others.

konklone:
> Running the stop command in the fabfile stops all the webhooks, but start only starts the one you specify.

Ugh, that's a bug. Thanks for noticing that.

mbland:
Right, I noticed that before, too. This is how it's _supposed_ to work:
```
help:      start               Start SCRIPT as a daemon
help:      stop                Stop the daemon SCRIPT
help:      stopall             Stop all running forever scripts
```

PR #796, opened 2015-05-01
Press page

melodykramer:

PR #863, opened 2015-05-27
New content for the home page

gboone:
Includes some front page copy for the homepage about our lines of business and hiring and some styling to support those sections.


Gemfile
```diff
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 gem 'rouge'
 gem 'bourbon'
 gem 'redcarpet'
-gem 'jekyll'
+gem 'jekyll', '3.0.0.pre.beta2'
```
konklone: Are you sure? =)

gboone: Yes.I think we have a better sense of what to watch out for, plus some rough tests that will ensure our feed comes out an xml file and `robots` comes out as a `txt` file

gboone:
This PR now includes and closes #796.

PR #864, opened 2015-05-27
Dan Brown recap and images

awfrancisco:


gboone:
Looks good from here. Let's take a look on staging.

PR #865, opened 2015-05-27
Publish: "Dan Brown: The art of conflict in design process"

gboone:


gboone:
This PR now also includes the new homepage content as seen on Staging.

PR #867, opened 2015-05-27
Spacing/styling fix for home page

gboone:
The spacing and styling was off for the "our work" section of the home page. This was due to a mis-reading and misapplication of classes and thus also css styles. This also adds new microformats that actually make sense.


elainekamlley:
:+1:

PR #871, opened 2015-05-27
18F guides post, subhead format and headline style updates,

awfrancisco:
Changed headline style to lower case on recent blog posts
Updated subheads to H2 where they weren't
18F guides image
18F guides post to be published Thursday, May 28.


gboone:
Looks good for a mornin' merge.

PR #874, opened 2015-05-28
semi-urgent typo

gboone:
Some idiot (me) put sites.18f.gov instead of pages.18f.gov


gboone:
Thanks to @NoahKunin and @konklone

PR #878, opened 2015-05-29
renamed news to blog

### LINKED ISSUE #356: redirect /blog to /news
plz

elainekamlley:
Changed the following areas from `/news` to `/blog`:

- _config.yml
- global nav button
- took out `back to all posts` button in blog post layout
- renamed news folder to blog folder

closes #356

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -55,7 +55,7 @@ GEM
       jekyll (>= 2.0)
     json (1.8.2)
     kramdown (1.7.0)
-    liquid (3.0.2)
+    liquid (3.0.3)
     listen (2.10.0)
       celluloid (~> 0.16.0)
       rb-fsevent (>= 0.9.3)
--- a/_config.yml
+++ b/_config.yml
@@ -44,9 +44,9 @@ url: https://18f.gsa.gov
 title: 18F Digital Services Delivery
 description: "18F builds effective, user-centric digital services focused on the interaction between government and the people and businesses it serves."
 
-# blog archive at /news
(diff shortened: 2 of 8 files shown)


konklone:
Huge :+1: for this change. How will redirects be handled?

gboone:
We were just talking about that. I'm thinking an nginx rewrite rule. Something like:
```
rewrite ^(/news/.*)/blog/(.*)\..*$ permanent;
```

konklone:
That will totally work (haven't verified the syntax), but the less in nginx, the better. The more we put in there, the harder it is to leave nginx. 

Can we do this with Jekyll directly? Thankfully, we never included `/news` or `/blog` in our actual post URLs, e.g. `https://18f.gsa.gov/2015/02/09/the-first-gov-domains-hardcoded-into-your-browser-as-all-https/`, and even the tags page doesn't have `/news` in it.

So we only have to change `/news/`, from what I can tell. That's a `<meta refresh>` tag away.

gboone:
We could do that. I'm told doing redirects server-side is preferred when possible, but as long as search engines get the 301 redirect I don't know that it totally matters. The only pages that are affected are `/news` and `/news/page/*`

konklone:
You're right, server-side is preferred, though a small number of `<meta refresh>~ tags with a refresh of `0` will be treated as a 301 redirect by Google and Yahoo, and shouldn't cause us to get flagged as spam (having reams of meta refreshes can be a bad signal). In my opinion, it's a better balance in terms of the flexibility of the site.

PR #883, opened 2015-06-02
Merge team.json data into  _team/* front matter

mbland:
As discussed in #18f-site-dev, this selects fields from the imported `_data/team.json` to merge into the front matter of `_team/*.md` files.

cc: @gboone

### DIFF
--- a/_data/import-public.rb
+++ b/_data/import-public.rb
@@ -5,7 +5,10 @@
 # Author: Mike Bland (michael.bland@gsa.gov)
 # Date:   2014-12-22
 
+require 'hash-joiner'
+require 'json'
 require 'open-uri'
+require 'safe_yaml'
 
 DATA_DIR = File.dirname __FILE__
 DATA_BASEURL = 'https://18f.gsa.gov/hub/api/'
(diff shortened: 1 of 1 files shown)


mbland:
Note that I didn't actually commit the changes generated by this script; figured that can happen in a follow-up PR.

_data/import-public.rb
```diff
@@ -17,3 +19,17 @@
     end
   end
 end
+
+data = JSON.load File.read(File.join(DATA_DIR, 'team.json'))
+fm_rexexp = Regexp.new('^---.*---$', Regexp::MULTILINE)
+
+Dir[File.join [DATA_DIR] + %w(.. _team *.md)].each do |member_file|
+  name = File.basename(member_file)[0..-4] # Trim .md extension
```
dhcole: @mbland is it worth splitting on `.` instead of trimming the string? Jekyll supports other file extensions like `.markdown` or `.textile`

mbland: Yep! Though, splitting on `.` alone may prove reliable. In fact, just noticed that `File.basename` takes a suffix argument that does the job. Updated.

Also, FYI, `Jekyll::YAML_FRONT_MATTER_REGEXP` is defined in `lib/jekyll/document.rb` and duplicated in `Jekyll::Convertible::read_yaml` (`lib/jekyll/convertible.rb`). I'll update this to use that define regexp.

mbland:
So if this seems like a tool, or at least an approach, y'all'd like to take, two things:

- This currently requires using the existing JSON import from the Hub, not the hand-crafted update that turned the data into a list of objects. The script _could_ be updated to both process the Hub-imported data, and then munge it into this new format, if you decide that's what you want.
- The `props` array will need to be updated with all the fields you'd like to filter and add to the front matter of each page.

I'm really intrigued by this model of using the imported API data to inject front matter into individual pages.

gboone:
Yeah, not sure how I feel about binding it with the import data script or forcibly overwriting every team member's front matter every time but I think this is a good approach to start. I had started working on this problem as a jekyll plugin to create a `migrate` command that would let us target individual bios. So, if we knew the data in `team.json` was correct we could overwrite it.

On that note, though, I'm beginning to think it might not be necessary to keep `team.json` committed to the repo and instead read it out of the API whenever we need it.

mbland:
Closing this since it seems we're moving towards the Team API for a lot of this stuff, and I've got more robust code (to be added to the [`team_api` gem](https://github.com/18F/team_api) or somewhere) for updating a doc's front matter.

PR #885, opened 2015-06-02
Toward a better bio template

gboone:
This PR will include a new layout for team bios that will hopefully be reusable by other interior pages


gboone:
This is not ready for merge, it is an in-progress pull request.

jenniferthibault:
possible quick todo's for mvp:
- [ ] sidebar typography styling: category headers should be consistent
- [ ] move photo into sidebar column so that there is padding between the end of the blue bar and the photo
- [ ] add back button to list of people that jumps you to anchored position
screen shot 2015-06-09 at 12 19 22 pm

maybe not possible for mvp, but would improve the template
- [ ] image hover effect on all team page 
screen shot 2015-06-09 at 12 22 07 pm
- [ ] make the blue bar on the individual bio page span the whole width of the page

gboone:
TL;DR I think this is just about ready to go. 

I spent a lot of time this afternoon following a short pairing session with @jenniferthibault touching things up along the grid and nailing down font consistencies across the template. I think this is ready for a thorough review later this week. @elainekamlley we can chat about when we want to do that since I don't think we'll have time for that an all the other things during the mid-sprint review tomorrow. Plus :airplane:.

Before we get too excited about merging this, I think we want to encourage people to add (or directly reach out and do it for them) at least these data:
- [ ] team
- [ ] role
- [ ] location (displayed as the full name of the state, unlike in other places and previous iterations of team data)

Because we don't have the OGC clearance to add personal identifiers like blogs and twitter accounts, we're in no hurry to collect those just yet. But if we can it'll save us time later. Personal profiles can be updated in `_team/` and each team member _should_ have a markdown file. We're also committing ourselves to creating a new one of those and gathering all those data every time someone joins the team. Finally, in order to make the profiles build in production we'll need to edit the `_config-deploy.yml` file and remove the `team` section from the collections. We can merge this and get the new designs up on staging without that change, though.

As for the "beyond this sprint" things:
* I experimented with spanning the blue bar all the way across the screen and have two thoughts on it 
    1. It started looking a bit stripey (great design term, I know), kind of like how on giant screens it looks with the red bars stretching all the way. We might consider other alternatives to wider screen displays: a border around the main container and make the background color for `<body>` a shade of gray to give it some definition, for example.
    2. There will be a bit of re-working some of this CSS in order to make that design play across all viewport sizes.

* The hover effect I'm sure can be done but will need more time and attention (or maybe just someone much better at debugging SASS than I am. :shrug: )

gboone:
This PR is go for launch.

PR #905, opened 2015-06-08
Downgrades Jekyll

gboone:
Hoping this will fix the issue we're having on staging. Want to get Travis to take a look and make sure all's well before merging this one.


gboone:
merging #904 into this to give that shot

PR #908, opened 2015-06-08
Freezes redcarpet at v 3.2.3

gboone:
@konklone will this ensure we don't accidentally upgrade recarpet again?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 gem 'rouge'
 gem 'bourbon'
-gem 'redcarpet'
+gem 'redcarpet', '3.2.3'
 gem 'jekyll', '3.0.0.pre.beta2'
 
 group :jekyll_plugins do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -98,6 +98,9 @@ DEPENDENCIES
   jekyll_pages_api
   jemoji
   pry
-  redcarpet
+  redcarpet (= 3.2.3)
   rouge
   safe_yaml
+
+BUNDLED WITH
+   1.10.3


konklone:
Should, but we have a Travis error.

gboone:
Yeah, and it's the same one we had before...

gboone:
Wahoo, thanks Travis!

monfresh:
What was the reasoning behind locking the version? redcarpet 3.2.3 has a known security vulnerability. I went through the git history but none of the commit messages explained why these changes were made.

konklone:
As I recall, there was a regression of some sort when it upgraded from 3.2.3 to 3.3.0, but I can't find any details of what the regression was. It didn't come up because of a security vulnerability notification, and it would be worth us getting notices in place via [Gemnasium](https://gemnasium.com/) or something similar to avoid this going forward.

monfresh:
I have a Gemnasium alert. That's how I ended up here :) I was planning on upgrading redcarpet but saw that it was locked and couldn't find any documentation in the Git history. Is it safe to assume that the test suite is robust enough to catch any issues if the new version is incompatible?

gboone:
I recently tried updating this again and didn't make much progress. Any newer version of redcarpet with our current version of ruby fails to build. I think upgrading redcarpet might involve moving to a newer version of ruby.

gboone:
Well, whaddya know. I just tried updating redcarpet again and it seems to have worked. The only thing our test suite won't catch is any change to how redcarpet interprets markdown syntax. I'm going to open a pull request @monfresh and we can jam on that together?

gboone:
See #1791

PR #927, opened 2015-06-11
Travis will fail until morale improves (or until we fix the frontmatter)

gboone:
These tests fail! And that's good! Because we have missing frontmatter all over the place and we need to fix it so we don't have funny looking posts in production. I propose we leave this open until we've fixed all the frontmatter problems and then merge away! (Also I should comment this code a bit.)

The frontmatter tests will detect missing or incorrect frontmatter attached to posts. For example, we expect all posts to have an `authors` key in the frontmatter (even if there's more than one author) but sometimes we put `author` instead. This will catch that problem before it hits production.


deploy/tests/frontmatter.rb
```diff
@@ -0,0 +1,74 @@
+require 'yaml'
+def loadconfig()
+	config = YAML.load_file('_config.yml')
+end
+@config = loadconfig
```
konklone: Code style note, this is (IMO) a cleaner way of expressing this:

```ruby
def config
	@config ||= YAML.load_file '_config.yml'
end
```

And then subsequent references can be made to `config` instead of `@config`.

This gives you the same effect -- the config file is only read in once, and is cached in memory for subsequent calls. So you can continue to reference `config`, and it hardly matters whether it's a method or variable.

I also removed the empty parens, which in most Ruby code I've seen/written aren't used for methods which take no parameters.

deploy/tests/frontmatter.rb
```diff
@@ -0,0 +1,74 @@
...
+		if data.has_key?(key)
+			require 'pry'; binding.pry
+			return data[key].class == type
+		end
+	end
+end
+
+def check_rules(data, schema, file)
+	YAML.load_file('deploy/tests/schema/rules.yml')
+end
```
konklone: You can use a similar construction to my notes from above.

deploy/tests/frontmatter.rb
gboone: This is good to know, @konklone thanks for the tip. I was looking for a cleaner way of expressing this.

mbland:
The schema stuff exactly the kind of thing I was hoping to do a long time ago with [18F/data-validator](https://github.com/18F/data-validator), but I like how much easier this format reads.

gboone:
Might take some time migrating this over to 18F/data-validator now that others are paying attention.

Here's what is working right now:
- Matches keys in a collection document's front matter with the keys in the schema file
- For each key, matches the value "type" for if an array or string was expected

gboone:
To merge this:
- [ ] Merge #1057 
- [ ] Merge #1059 
- [ ] Make sure Travis is green

gboone:
As of 2259420 we're running the tests out of a newly-cut gem instead of a ruby script in the deploy branch. You can run these tests yourself by doing `bundle install` and then `jekyll test`. These test _should_ go green once we merge #1059.

gboone:
Going to merge #1059 into this branch to close it. Then we can merge this one and make Travis go green!

elainekamlley:
Woop!

PR #930, opened 2015-06-12
WIP updating footer content

elainekamlley:
This is a WIP, please do not merge! 

I have added new links that navigate to newly added content or 18F projects. Working to get the footer  code closer to the footer in the homepage mock ups @jenniferthibault made: [Here](https://drive.google.com/a/gsa.gov/folderview?id=0B1PRnuw_hqg1flV4bjhWc0hEb25QOWNMUGhTR2s3RzRma3BiMVZaN1FKUUpGdENsYTJsMXc&usp=sharing)

Still need to: 
- [x] Adjust logos as per mock up. (If there is time in this sprint) 
- [x] fix responsiveness and spacing. It does this not so great thing: 
screen shot 2015-06-12 at 3 09 33 pm
- [ ] Get feedback from site team on order of items. Still need to find a home for blog. Suggestions? 
screen shot 2015-06-12 at 3 11 37 pm

Would love any feedback or direction @awfrancisco @emileighoutlaw @kategarklavs @jenniferthibault 
Also any code review from @gboone @msecret


_includes/footer.html
```diff
@@ -3,21 +3,27 @@
     <div>
       <img src="/assets/images/logo-gsa.png" alt="GSA logo">
     </div>
-    <div itemscope itemtype="http://schema.org/ContactPoint">
-      <h1>Contact Us</h1>
-      <a href="mailto:18f@gsa.gov" itemprop="email"><i class="fa fa-envelope"></i> 18F@gsa.gov</a>
+    <div>
+      <h1>Work</h1>
+      <a href="/consulting">Consulting</a>
```
msecret: The <a> tags should have `title` tags for accessibility. They should describe what the link is liking to, in the case, "consulting" would probably be fine.

_includes/footer.html
```diff
@@ -3,21 +3,27 @@
...
-      <a href="https://github.com/18F"><i class="fa fa-github"></i> github.com/18F</a>
-      <br/>
-      <a href="https://github.com/presidential-innovation-fellows"><i class="fa fa-github"></i> github.com/presidential-innovation-fellows</a>
-      <br/>
-      <a href="https://github.com/18F/18f.gsa.gov/edit/staging/{{ page.path }}">Edit this page</a>
-      <br/>
-      <a href="/developer/">Developer resources</a>
+      <h1> People </h1>
+      <a href="/team">Team</a>
+    </br>
```
msecret: Although I technically think using a <br> to move the links to new lines is fine, a better approach to make things work responsively would be to use CSS. So rather the using <br> tags between each tags, you can use a class like `.footer_nav` and use display: block for each link

```css
.footer_nav a {
  display: block
}
```

Again, I'm not sure this is necessary for based on what this should do at different screen sizes, so consider that when making a change.

_includes/footer.html
```diff
@@ -3,21 +3,27 @@
...
+      <a href="https://pages.18f.gov/joining-18f/">Join us</a>
+    </div>
+    <div>
```
msecret: Another idea I had would be to use <section> html5 tags for these. Not strictly necessary, but may make more semantic sense.

assets/_sass/_custom.scss
```diff
@@ -1184,7 +1184,7 @@ $tiniest: new-breakpoint(max-width 390px);
 	}
 	padding-bottom: 20px;
 	h1 {
-		font-size: 1.2em;
+		font-size: 1em;
```
msecret: Nice job using em!

awfrancisco:
I would vote to swap the order of Consulting and Delivery under the Work section to match the order of the Our Work section above.

assets/_sass/_custom.scss
```diff
@@ -1184,7 +1184,7 @@ $tiniest: new-breakpoint(max-width 390px);
 	}
```
msecret: The CSS you edited all looks fine here. I do see an existing problem of deeply nesting. It might help long-term style maintenance by keeping the css modular. So you would have one _footer.scss file to keep this stuff in like this:

```css
.footer
.footer-container
.footer-link
...
``` 

This isn't necessary now, but something to consider if you do refactoring of styles in the future.

elainekamlley:
Thanks for all the great feedback @msecret! I incorporated much of your feedback with adding `titles` as well as adding a inline-block to `a` tags which works for now in terms of responsiveness. @gboone and I will have a performance sprint soon and I want to think more about making the CSS more modular to include the top nav bar as well. @gboone want to take a look? If everything looks okay, want to see it on staging? 

@awfrancisco switched it up! Any other suggestions for nav order?

gboone:
Taking a look this morning. Thanks for the feedback, @msecret

elainekamlley:
Closing this PR, work was finished and merged on #1066

PR #938, opened 2015-06-16
List type fix

gboone:
List styles, spacing, and margins should be standardized across blog posts _and_ the `bare` template now.


meiqimichelle:
Niiiiice :)

PR #945, opened 2015-06-17
Adds DAP agency abbreviation as required

philipashlock:
As per requirements in https://www.digitalgov.gov/files/2015/02/GSA-DAP-UA-Code-Quick-Guide-15-01-30-v1-02_mvf.pdf


konklone:
:+1: We're adding `subagency=OCSIT` to analytics.usa.gov's snippet too, FWIW.

Issue #542, opened 2015-02-19
How to Use Github and the Terminal: A Guide


melodykramer:
The tutorial is available on our blog: https://18f.gsa.gov/2015/03/03/how-to-use-github-and-the-terminal-a-guide/ come back here to discuss how we can improve it!

gboone:
Note: Anyone who is not an 18F employee will not be able to push a new branch to this GitHub repository. If you try you will get an error.

seanherron:
This is really fantastic! As a suggestion, maybe add screenshots or inline GIFs of you doing various command line activities? [LICEcap](http://www.cockos.com/licecap/) is my tool of choice. Then you can do cool things like this:

![image](https://cloud.githubusercontent.com/assets/297696/6276648/5cca0f20-b856-11e4-848b-edfd88d734d4.png)

melodykramer:
SO COOL. Will add that. @seanherron. Thanks!

stvnrlly:
I haven't been able to try it yet for inclusion in the Code for DC guide, but Git for Windows supposedly provides an interface with Mac/Linux commands, which would make this guide universally applicable: https://msysgit.github.io/

melodykramer:
That's a great idea @stvnrlly -- I also love your guide: https://github.com/codefordc/guides/blob/master/learn-github/learn-github-osx.md

stvnrlly:
Thanks! I'm going to see what I can "borrow" from this. We also have materials for other stuff you cover here:

[Installing Jekyll](https://github.com/codefordc/codefordc-2.0/blob/master/README.md)
[Submitting a post](https://github.com/codefordc/codefordc-2.0/blob/master/_resources/submitapost.md)

Here are a few small things:

- I don't believe Homebrew will work on a Linux machine. The package manager would vary from distro to distro, but the instructions should be the same with that manager's command plugged in.
- I don't see `brew` actually used here. I had used it to install Git: that appears to be redundant, but I must have had some reason for it...
- You may want to recommend installing RVM no matter what. Ruby's use of the path for commands can be really confusing if you don't know what's going on, and RVM avoids this. (I see that a lot is happening in the `./go` command, so this may be unnecessary)
- Headers would be nice to break things up and create link points.
- The YAML front matter example implies that you should type `--- on line 1` rather than `---` on line 1. It may be clearer to just say "Don't forget the triple dashes."

gboone:
Thanks for your feedback @stvnrlly!

[...]

PR #947, opened 2015-06-19
remove google code, it's EOL

jbjonesjr:
Since you can no longer [add new projects to google code](http://google-opensource.blogspot.com/2015/03/farewell-to-google-code.html), it doesn't make sense to list it as a GitHub alternative.


NoahKunin:
Thanks @jbjonesjr! 

Nice catch - quick FYI for others as well, the entire Google Code is coming down in Jan 2016. Merging!

jbjonesjr:
/plugmode Google Code provides a [1 button migration option](https://code.google.com/export-to-github/) to get your code to GitHub to continue functionality. /plugmode

gboone:
That's great advice, @jbjonesjr. We don't have any code hosted there but you might let others know in the discussion at #542. Thanks again!

PR #948, opened 2015-06-19
Correcting typo in excerpt

jameshupp:
h/t to my grandmother, who caught it.

PR #949, opened 2015-06-19
DAP fix, google code deprecation, and typo correction

gboone:
Deploys #947 #948 and #949 to production.


_includes/scripts.html
```diff
@@ -50,4 +50,4 @@
 </script>
 
 <!-- Digital Analytics Program roll-up, see https://analytics.usa.gov for data -->
-<script id="_fed_an_ua_tag" src="https://analytics.usa.gov/dap/dap.min.js"></script>
+<script id="_fed_an_ua_tag" src="https://analytics.usa.gov/dap/dap.min.js?agency=GSA"></script>
```
konklone: We could do `subagency=18F` here, if we wanted.

gboone: is that in addition to `agency`?

PR #953, opened 2015-06-22
text changes

awfrancisco:
Updated headline and text to be more explanatory. Extended timeout from 20 to 30 seconds.


gboone:
Looks good!

PR #967, opened 2015-06-29
link to hiring site from job-related blog posts

afeld:
Occurred to me that these two posts are still coming up in search results, so we should direct people to the up-to-date information. Totally open to other ideas about how to word/format it.


gboone:
Good idea. Thanks @afeld

PR #975, opened 2015-06-30
Press kit style fixes

gboone:
augments #974 with accessibility fixes and style changes


gboone:
Also adds a `min-height` to the headline to support smaller screens. This wasn't a noticeable issue before because the images on team profiles expanded the box sufficiently.

PR #978, opened 2015-07-01
Applies the new standard template to /consulting.

gboone:
In theory this template is ready for all the pages but we'll want to proof them to look for design inconsistencies before going primetime.

Changes from the /press and /team page commits: you can now specify a "subtitle" on a page and it will show up in the second line of the "headline" for any document using the layout "page."

Outstanding changes remaining: create a style for posts to use the headline. This should be relatively straightforward at this point.

cc: @awfrancisco @jenniferthibault @elainekamlley

### DIFF
--- a/_includes/consulting-side.html
+++ b/_includes/consulting-side.html
@@ -1,0 +1,18 @@
+<aside>
+	<h2 id="recent-projects">Recent projects</h2>
+	<ul>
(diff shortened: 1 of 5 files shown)


gboone:
Going to say this has been out long enough that the author should revise and resubmit.

PR #983, opened 2015-07-02
Further rewrote my bio a bit

konklone:
I removed the third-person paragraph, removed my affiliation with the (now nonexistent) Policy Guild, and reworded things slightly.


gboone:
thanks @konklone

PR #984, opened 2015-07-02
Update lindsay.md

LindsayYoung:
adding some links


elainekamlley:
Hey @LindsayYoung, for the bios we can link to the work you did here at 18F (FEC API) but can't have a direct link out to previous work. We can mention it but can link out to it.

PR #989, opened 2015-07-02
fixed some typos/language issues, changed role to be consistent with others at 18F

onezerojeremy:
- typo: "leared" -> "learned" (though i do like the idea of leering at how government works)
- cleaned up some phrasing in the quote 
- role  "UX Researcher" (was only one here, not how I self ID) -> "User Experience Designer" (others here have this title, is one of the things ppl call me)


gboone:
Thanks @onezerojeremy

PR #990, opened 2015-07-02
more folks

elainekamlley:
Adding: James Scott, Will Sullivan, and Zac Cohn
Pic: Jacob Harris


gboone:
Thanks @elainekamlley @jcscottiii @Journerdism @zacharycohn @catherinedevlin and @harrisj

PR #1000, opened 2015-07-06
Takes out the back link on /tags archives

gboone:
We already have a link to /blog in the global nav menu. Will update this if I find more that we missed but this might be the last one.


elainekamlley:
1000 :boom: :exclamation:

Issue #115, opened 2014-07-18
how to handle redirects


gbinal:
When 18f.gsa.gov/developer goes live, I'd like for 18f.gsa.gov/developers to redirect to it.  It's a small thing but a good best practice.  I'm happy just creating a simple html redirect at /developers.html for now.  Any preferences otherwise?

PR #1016, opened 2015-07-09
adding a redirect for /developers

gbinal:
It's a small but simple best practice that you want both `x.gov/developer` and `x.gov/developers` to arrive at the developer hub since it's easy for folks to misremember which it is and get a 404 if they are wrong.  

Since this was [discussed](https://github.com/18F/18f.gsa.gov/issues/115) a year ago, we've updated our practices and [this method of offering redirects is allowable](https://18f.slack.com/archives/g-accessibility/p1432221216000214).  

This pull request should result in `18f.gsa.gov/developers` redirecting to `18f.gsa.gov/developer` instead of 404ing.  

https://18f.gsa.gov/developer/

### DIFF
--- a/pages/developers.md
+++ b/pages/developers.md
@@ -1,0 +1,8 @@
+---
+layout: bare
+permalink: /developers/
+---
+
+<META http-equiv="refresh" content="0;URL=https://18f.gsa.gov/developer/">
+
+This page is redirecting to the [18F Developer Hub].  If the page doesn't not redirect you automatically, please <a href="https://18f.gsa.gov/developer/">click here</a>.  


gboone:
Cool. For future redirecting you can make use of the `jekyll-redirect-to` plugin and accomplish this same thing with just frontmatter. [Example: https://github.com/18F/18f.gsa.gov/blob/staging/pages/tags/18f%20events.md](https://github.com/18F/18f.gsa.gov/blob/staging/pages/tags/18f%20events.md)

gbinal:
Ah!  Good to know, thank you.  

If this has any problems, i can change it over to that anytime.

PR #1033, opened 2015-07-15
Update _typography.scss and _button.scss

andrewmaier:
Only apply antialiasing to high-resolution displays.


andrewmaier:
@gboone

PR #1041, opened 2015-07-16
Fixed a couple of typoes and redundancies

esgoodman:
I was just browsing and found some stuff.


elainekamlley:
Thanks @esgoodman !

PR #1048, opened 2015-07-22
Hotfix for broken links in re: to Agile BPA and CIO Council presentation

NoahKunin:


konklone:
This also includes an image fix, and an email link fix, both of which look :+1: as well.

PR #1049, opened 2015-07-22
Jeremia's photo and correct spelling

jeremiak:
My name was misspelled and I didn't have a photo, but I've corrected those in this PR :)


gboone:
Looks like we spelled it right before immediately spelling it wrong. Sorry about that! Thanks for catching it, though.

PR #1054, opened 2015-07-26
template page already puts a period after cta

drewp:
https://18f.gsa.gov/ shows "doing it.." instead of "doing it."


konklone:
So it does:

image

Thanks, @drewp!

PR #1081, opened 2015-08-06
Does a 1080

gboone:
Deploys #1080 to production fixing a typo caught by @bsweger. (Thanks, @bsweger!)
Also, does a 1080.


cmc333333:
<3 @gboone

konklone:
That was exhilarating

bsweger:
I am so confused.

PR #1103, opened 2015-08-18
Updating Julia Elman bio with city, twitter and github handles.

juliaelman:
Small pull request with a few bits of update information.


gboone:
Thanks @juliaelman!

PR #1107, opened 2015-08-20
Empty author fix

gboone:
Fixes a bug where authors would get a URL even if not on our team. Also lifts a ton of logic out of the templates and into the plugin where it can be done more efficiently. This made it possible to add a custom URL to guest authors. Add a `url` key to the listing in `authors.yml` and the custom URL should appear in a blog post.

This creates a possible UX inconsistency on blog post pages, though, where most of the time a link on a post will take you to a profile page where you can learn more about the author and find other posts written by that person _except_ when for guest authors. In case of a guest author you might be taken away from 18f.gsa.gov. We should be careful to make sure the kind of content on pages linked to these cases is a profile or other page where you can learn more about the person.

Also adds a bio for garren and closes 18F/blog-drafts#285


gboone:
cc @18f/blog

gboone:
Paging @18F/blog

PR #1117, opened 2015-08-26
Remove "independent agency" callouts

konklone:
GSA is an "independent establishment", but not an independent agency in the sense as the FTC or FCC -- unlike the FTC or FCC Commissioners, the Administrator serves at the pleasure of the President. There's some confusion about this, in part because of [GAO's ambiguous mixing](http://www.gao.gov/browse/agency/Independent_Agencies) of "independent establishments" and "independent regulatory agencies" as "independent agencies". Better to just leave it out entirely.


gboone:
Going to push an edit to change some `.  ` to `. `

konklone:
I don't know how those got there. I would never do that.

PR #1144, opened 2015-09-03
every kid in a park post and images

awfrancisco:


gboone:
Looks good from here, let's check it out on Staging

PR #1157, opened 2015-09-08
Moves DOL handbook post to staging

awfrancisco:
Add DOL post images
Adds jtag as an author
Fixes a tag on the Lego post


gboone:
This is for production tomorrow, right?

awfrancisco:
Yup.

PR #1168, opened 2015-09-14
Added alt attribute to image in .gov domain post

adelevie:


konklone:
Thank you! I need to be more mindful of this, and it'd be great to build in a little linting tool into the site.

adelevie:
FWIW, I used the HMTL Code Sniffer Chrome extension. 

`pa11y localhost:whatever/whatever` will also get you the linting.

PR #1170, opened 2015-09-14
Use Team API instead of Hub API

### LINKED ISSUE #1169: Use Team API Instead of Deprecated Hub API
https://github.com/18F/18f.gsa.gov/blob/staging/_data/import-public.rb#L11

jessieay:
* Fixes #1169
* `pif_projects` `pif_team` and `projects` json files were not in use


jessieay:
@gboone would love a review/merge when you have a chance - after this is merged, next up is importing new records as `md` files when they are in the API  but dont have a file in this project

PR #1171, opened 2015-09-14
Add markdown files for new team members

jessieay:
* If no existing file for a team member, add basic info to markdown
* Variation of https://github.com/18F/18f.gsa.gov/pull/883


jessieay:
The good news is that this script works! The bad news is that it introduces a few dupes (See: valdiviezo). This is a problem with the "name" attribute not being used consistently for file names. 

Happy to just re-name files that are not already the same as the "name" attribute for team members in the API - eg: Victor is "valdiviezo" in the team API but has a markdown file called "victorvaldiviezo" currently in this repo. 

Should I go ahead and rename existing dupes or is this a symptom of a larger consistency problem we need to address?

_data/import-public.rb
```diff
@@ -9,12 +9,60 @@
 require 'json'
 
 DATA_DIR = File.dirname __FILE__
+TEAM_DIR = "_team"
 TEAM_API_BASE_URL = 'https://team-api.18f.gov/public/api/'
+TEAM_ENDPOINT = 'team'
+MEMBER_PROPERTIES = [
```
mbland: You can define an Array of string constants using the `%w(foo bar baz)` style. (Which I learned from [Rubocop](http://batsov.com/rubocop/) on other projects.)

mbland:
Yes, that is an issue, that we need better coordination of 18F user IDs across properties (perhaps the Team API being up-and-running will help motivate this). However, there's also the issue that we haven't actually archived former members, such as `greg` and `majma`, per 18F/data-private#70. It seems it's time we prioritized that.

gboone:
Yeah, we'll have to fix the archived team members problem before we merge this. There are seven profiles that will generate now that shouldn't. That was actually the primary reason we we waited to adopt the Team API. As I mentioned in 18F/data-private#70 we have no use for former team members on 18f.gsa.gov other than blog post bylines but we've solved that problem with a separate authors.yml file. I'm a vote in favor of removing anyone who is not a current team member from the Team API.

_data/import-public.rb
jessieay: is that your preferred style? I tend to stay away from `%w` because I sometimes forget the difference between it and `%i` and all the other fancy shorthands in Ruby. This would be a good convo to have in https://github.com/18F/engineering-best-practices

gboone:
As for the duplicates, @jessieay I'd say we can delete the existing `_team` file and keep the one that matches the Team API data. This is a parallel movement problem we're hopefully solving by consolidating with Team API.

jessieay:
@gboone alright, moved all md files and images so that they are using the Team API `name` field as the file name.

Do you see any new files for people who should not be included? Manually removed the two 18F alums identified in https://github.com/18F/data-private/issues/70

jessieay:
cc @mbland

gboone:
The others are:
* greg
* majma
* michelle
* robert
* sean

jessieay:
@gboone alrighty, removed those. Next up: end dates in Team API! Then this script can just be run without us having to worry about deleting alums

jessieay:
@gboone @mbland this is updated and using `name` fields that are currently in the Team API. Still had to manually remove the 7 alums, but we can automate that once `end_date` is in the Team API

gboone:
@jessieay sorry I've been a bit out of touch on this PR. Is this ready to merge or are we waiting for something else?

jessieay:
@gboone this PR is kind of ready....it does need to be rebased, which I can take care of today.

The larger issue with the script it contains is that it pulls *all* people from the Team API, including people who are no longer with 18F.

Once we have employment dates in the Team API, this PR can be updated to only pull in current employees.

@mbland what's the latest on getting end dates into the Team API?

gboone:
> The larger issue with the script it contains is that it pulls all people from the Team API, including people who are no longer with 18F.

I think I have a part time fix for this. I added all the former team members to the `gitignore` file. In theory we could patch the script to auto-delete those files, but as long as we don't 1) run the script in production deployments and 2) forcibly version control those files I think this is good to merge. We can update with end dates when they're there.

mbland:
Still open as 18F/team-api.18f.gov#123. We may prioritize this for the next sprint, which we'll be planning in the next couple hours.

jessieay:
Sounds like a good plan, @gboone -- what would you like me to do with this PR in the short term? Looks like you updated recently but there are still some merge conflicts. Would you like to handle those or should I? (also a lot of Hound comments to deal with)

jessieay:
@gboone never too early to use Hound! Just fixed merge conflicts, updated styles, squashed, and pushed. Ready for final review / merge once CI passes

_data/import-public.rb
```diff
@@ -1,20 +1,65 @@
...
 end
+
+def add_files_for_new_team_members
+  team_json_data = JSON.parse(team_data)["results"]
+
+  team_json_data.each do |team_member_data|
+    team_member_name = team_member_data["name"]
+    team_member_file_path = File.join(TEAM_DIR, "#{team_member_name}.md")
+
+    unless File.exist?(team_member_file_path)
```
jessieay: disagree, Hound

_data/import-public.rb
```diff
@@ -1,20 +1,65 @@
...
-require 'json'
+require "open-uri"
+require "json"
 
 DATA_DIR = File.dirname __FILE__
-TEAM_API_BASE_URL = 'https://team-api.18f.gov/public/api/'
+TEAM_DIR = "_team"
+TEAM_API_BASE_URL = "https://team-api.18f.gov/public/api/"
+TEAM_ENDPOINT = "team"
+MEMBER_PROPERTIES = [
```
jessieay: disagree with this

gboone:
Looks good, thanks! Let me know if you have some time to help us configure Hound or have thoughts on the matter. I'm noting your disagreements with Hound above.

jessieay:
@gboone glad someone witnessed my argument w a bot! 

Would love to work on some rubocop rules for 18F that make sense to everyone. For now, letting the #wg-code-style-guide crew get started on that :)

PR #1173, opened 2015-09-15
Publishes: Federalist platform post + Team API

gboone:
Pulls team member info from the Team API with @jessieay's contributions in #1170. HUGE THANKS!
Also publishes new blog post staged in #1172.
cc: @18F/blog


jessieay:
Do pull requests for merging staging into production usually get a code review? Since we code review before things go into staging and I consider that "master", I would default to not requiring code review to merge staging into production, assuming staging is not broken

gboone:
Yeah typically we don't do a code review before merging staging (our "master") to production. We do have a policy of the person who opens a pull request not being the same person who merges them, though. That's more a transparency issue than a "review" issue. We also hold deployments for timing, though. So, in this case, the blog post included in this PR was due to launch at 2. We _try_ to do production deployments as quickly as possible after merging a staging PR so they are as small/focused as possible. I'm not great at that, though.

PR #1174, opened 2015-09-15
Update 2015-09-15-federalist-platform-launch.md

batemapf:
@gboone ready for you.  basically took out anything that has to do with us charging for the product and saving people $ (which it will do, but we aren't ready to talk about yet).


gboone:
I don't know what's keeping Travis so busy. This looks good and will eat my hat if it somehow breaks the build.

PR #1202, opened 2015-09-26
Final edits

quepol:
Figured I'd save you some work, and have ported all the edits I just made to the Google Doc over here. Made the URL changes separate commits in case I missed a memo about the playbook URL changing. Otherwise, the link to the Standards was incorrect each time.

/cc @awfrancisco @gboone


gboone:
Thank you, @quepol!

PR #1211, opened 2015-10-02
Update profile to reflect new position

khandelwal:
I've now moved over to consulting. My 18F website profile should reflect that.


elainekamlley:
Whoo! Congrats again @khandelwal

PR #1219, opened 2015-10-06
Makes more accessible the HTML table in the post

gboone:
Thanks to @rtwell for noticing this table was not accessible.


gboone:
Basically what's happening here is the original table had a few problems. It was lacking a table heading and some of the `<a>` tags were closed without any content.

We fixed the table by adding a `thead` with a single `tr` and `th`s to label the columns.

The links were fixed by either removing a duplicitous anchor tag or simply moving the closing `</a>` to enclose some content.

gboone:
cc: @18F/blog

PR #1224, opened 2015-10-07
WORK IN PROGRESS: Jupyter-powered blogging

catherinedevlin:
Once we have a tmpnb cluster running, this should allow blog posts to include runnable live code.

Using O'Reilly's `Thebe` project, allow code samples with a `data-executable`
attribute to run live against a Project Jupyter server
(running with https://github.com/jupyter/tmpnb).
    
An example of how to use is in the (purely for demo purposes)
scratch post at examples/2015-10-07-how-we-blog-with-jupyter.md.
    
The Jupyter server is not yet spun up.  It is expected at jupyter.18f.gov:8000.
Until it is up, these scripts won't actually work.


NoahKunin:
So, my :heart: almost :boom:'d when I read "runnable live code...in blog posts". Super exciting, but also a a potential massive security boundary change. Hold on this until I can take a look to assess (going to try and figure it out early next week).

NoahKunin:
Ah fascinating - after a quick read-through, the threat model is actually in the *opposite direction*! `Thebe` is taking the code _from_ the static web page, and starting it into the kernel (why the **heck** did they call a process running inside of container on a server itself a kernel? Super-confusing) within Cloud Foundry. 

That's great, and the security model for CF should make this just work, but will want to still take a final look at it before merging to final `production`. Might go completely without saying, but it's absolutely critical we run this `over https`. Weird that `oreillymedia` even allows it to run on HTTP. Seems to just be inviting RCE otherwise. 

This is super cool - can't wait to see it on the site!

konklone:
This is _hugely cool_ and will continue to set us apart as a "government blog". 

@catherinedevlin, please just holler if you need any HTTPS assistance. I'm gonna guess that Jupyter is running as a Python daemon on port 8000, and isn't typically where you'd terminate TLS. I think ideally, we'd be deploying this to Cloud Foundry using a buildpack that has nginx sitting in front of Jupyter and proxying stuff back to it. We'd want this anyway for a bunch of helpful security/efficiency reasons. 

By using a Cloud Foundry buildpack with nginx in front of it, we'd also get the `*.18f.gov` cert in place for free, so you wouldn't need to worry about certs and stuff -- just getting the buildpack together.

If you don't want to make a buildpack (I couldn't find something ready-to-go out there), then you would at least want to use the fisma-ready AMI and nginx config to put nginx in front of Jupyter yourself.

gboone:
I think the open question is how exactly we'll get `Thebe` up and running on Cloud Foundry. I don't know nearly enough about CF or Docker to know exactly how it'll work but @catherinedevlin and @ozzyjohnson I think have spoken briefly about what the approach might be. I'll let one of them fill in the details.

gboone:
This seems to have stalled out, I'll leave the branch in place in case we want to re-start this work.

PR #1228, opened 2015-10-08
Publish: You shared, we listened — updates to the 18F Content Guide

gboone:
cc: @18f/blog @kategarklavs


kategarklavs:
And merging!

PR #1229, opened 2015-10-08
Adds Htmlproof + tests

gboone:
cc: @18f/blog @afeld 

This adds `html-proofer` to the entire set. Every time there's a pull request Travis will now run a few extra tests for us. They're called: "ImageCheck", "ScriptCheck", "LinkCheck".

1. ImageCheck inspects image tags for alt tags and makes sure the `src` exists (if it's internal). No more missing images on staging! HOORAY! It also seems to understand that a `figure` with an `image` and a `figcaption` means it's alright for the `alt` text to be missing.
2. ScriptCheck I think only checks that linked scripts exist.
3. LinkCheck checks _every internal link_ on the site. We can also set this up to check for broken links to external sites but it will greatly slow down the test. I haven't successfully gotten the external link checker working (@afeld have you?) but it might be worth setting up a nightly Travis run (can you do that in Travis? I'm not sure) to run with it enabled in the future.

Anyway, I think 3 is the big winner for me. How many times have we sent a post to staging but forgotten to include the images or had an incorrectly named image and then it vanishes on staging. So many! Never again!

This also fixes a whole bunch of posts and pages that were causing these tests to fail. Hooray better posts!


gboone:
Aaaand it failed. Better see what's up.

afeld:
> We can also set this up to check for broken links to external sites but it will greatly slow down the test.

Yeah, general good practice for continuous integration is to not have your builds rely on external services. For example, if we link to some random tutorial and they take their site down temporarily, our build breaks.

> I haven't successfully gotten the external link checker working

Do you mean "passing", or "running at all"?

> @afeld have you?

I think so...? It's been a while.

> it failed

This seems like a problem with the interaction of the `go` script and Bundler...paging @mbland!

afeld:
Does `./go ci_build` work locally?

mbland:
Fixed.

afeld:
:+1: from me!

gboone:
I figured it had something to do with bundler. :wink: Thanks for the assist, @mbland and the eyes @afeld. @18F/blog what say you?

awfrancisco:
I'm not qualified to inspect the code, but I'm :+1: on all the checks.

_layouts/press.html
```diff
@@ -9,12 +9,12 @@
...
   {% include headline.html %}
-  <div class="content" role="main" id="top main" itemscope itemprop="mainContentOfPage">
-    <a href="#toc" class="skips">Table of contents</a>
+  <div class="content" role="main" id="main" itemscope itemprop="mainContentOfPage">
+    <a href="#toc" id="top" class="skips">Table of contents</a>
     {{ content }}
     {% for item in site.press %}
       {% unless item.title == '18F Press' %}
-          <section>
+          <section id="">
```
konklone: Interesting, I wouldn't expect this to be best practice.

konklone:
How easy is it to extend this to do custom checks, like looking for `github.io`, or `http://en.wikipedia.org`, or other common stuff we run into?

konklone:
(Also it looks :+1: to me.)

afeld:
@konklone Very doable.

https://github.com/gjtorikian/html-proofer#configuration

_layouts/press.html
gboone: You might have caught a mistake on my part. I think I was supposed to get rid of that `id` altogether.

gboone:
@afeld 
> Do you mean "passing", or "running at all"?

I couldn't get it to run at all locally. I'd get two kinds of errors:
- `Couldn't resolve host name`
- `got a time out`

Any ideas? We can probably merge this without the external checker working since we don't want broken links to break our build anyway. It'd be nice to use this tool to check for broken links sometime in the future, though.

afeld:
> I'd get two kinds of errors...Any ideas?

Both of those are valid HTTP errors...try to `curl` the corresponding URL and you should see a similar message.

```
$ curl http://sngfgasdgsa.com/
curl: (6) Could not resolve host: sngfgasdgsa.com
```

Feel free to throw some time on my calendar if you want to look at it together.

gboone:
Oh they're totally valid errors. They're for hostnames like github.com and 18f.gsa.gov. When I curl them it's all good, I get request timeouts for `ping 18f.gsa.gov` but can't replicate the GitHub ones. Example:

```
error: External link https://18f.gsa.gov/2015/08/07/technical-debt-1/ failed: got a time out

$ curl -I https://18f.gsa.gov/2015/08/07/technical-debt-1/
HTTP/1.1 200 OK
Server: nginx/1.8.0
X-Government-Innovation: Disrupted

$ ping 18f.gsa.gov
Request timeout for icmp_seq 0
```
And for github.com
```
External link https://github.com/18F/18f.gsa.gov/edit/staging/_team/vdavez.md failed: 0 Couldn't resolve host name

$ curl -I https://github.com/18F/18f.gsa.gov/edit/staging/_team/vdavez.md
HTTP/1.1 302 Found
Server: GitHub.com
Date: Wed, 14 Oct 2015 12:38:28 GMT
Content-Type: text/html; charset=utf-8
Status: 302 Found

$ ping github.com
64 bytes from 192.30.252.128: icmp_seq=0 ttl=55 time=39.345 ms
```

PR #1235, opened 2015-10-09
Fixed caps on "PaaS", added link from GIF to source

mogul:
Saw on Hacker News and elsewhere that people wanted to know where the awesome GIF came from. Not very discoverable, but at least if they click they'll see it.

Note I made this edit via the GitHub web interface, and I'm not sure how Markdown will handle the  inside brackets... So please test before merging!


mogul:
Changed to use straight nested markdown rather than , and tested that this does render correctly in the GitHub inline-editor preview pane.

PR #1252, opened 2015-10-15
Update 2015-10-15-best-practices-for-distributed-teams.md

melodykramer:
Added an example to the "Use our words" paragraph to make it more clear for readers, based on a suggestion from Kate and a tweet online.


melodykramer:
@awfrancisco I made a slight change, based on a suggestion from Kate and a tweet.

PR #1253, opened 2015-10-15
Fixes responsive embeds

gboone:
OK, first attempt didn't quite work out. Here's another attempt at responsive embedding.


elainekamlley:
Looks pretty good to me on full screen to mobile but there is no left margin at the midway point at 656px. 

screen shot 2015-10-15 at 3 34 10 pm

gboone:
Yeah, I noticed that, too. Any ideas what might be the culprit there?

PR #1258, opened 2015-10-20
Fix a Typo

zeke:
s/team was lead/team was led/


gboone:
Good catch @zeke thanks!

PR #1280, opened 2015-10-30
More CSS Refactoring 

elainekamlley:
Did a few things to clean up the CSS. Was poking around and noticed that some styles were using the same styling across the site so move them in to `_base.scss`. Here's a list of things: 

- moved `p` & `h1` to site wide styling
- created `class=“h4”` for smaller `h1` tag in blog posts vs styling the `h1` tag itself
- move container padding to its own class `container-spaced` and removed `blog entry` and `service-entry` tags 
- increased `p` font size from 12 to 14 to be consistent in base

@gboone the big one I would love your opinion on is increasing the font size. @msecret and I noticed that the site uses 14px except for the blog which is slightly smaller. What are your thoughts on making it consistent throughout the site? Here are the differences: 

12px on the live site: 
screen shot 2015-10-30 at 9 47 10 am

14px on the branch: 
screen shot 2015-10-30 at 9 46 50 am

### DIFF
--- a/assets/_sass/_custom.scss
+++ b/assets/_sass/_custom.scss
@@ -43,8 +43,8 @@
 }
 
 .container-spaced {
-	padding-bottom: 40px;
-	padding-top: 40px;
+  padding-bottom: 40px;
+  padding-top: 40px;
 }
 
 .lead {
--- a/assets/_sass/_custom.scss
+++ b/assets/_sass/_custom.scss
@@ -20,7 +20,6 @@
 	}
 }
 
-
 // Extends
 //********************************************************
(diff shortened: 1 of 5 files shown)


gboone:
@elainekamlley looks like we've angered The Hound. How should we proceed? Configure Hound better?

elainekamlley:
@gboone taking a look to see if there is anything serious. But you know The Hound is always serious.

gboone:
This looks great, thanks for the ongoing CSS cleaning, @elainekamlley and @msecret it is much appreciated. :+1:

PR #1283, opened 2015-10-30
Jekyll upgrade

gboone:
This PR upgrades Jekyll to the newly polished version 3.0.0 gem. The big breaking change in the new version is that all collections are treated as equal "documents" in the site build. What does this mean? Check out the changes to `pages/team.html`. The biggest change there is switching `site.collections.team` to `site.team` when trying to find a team member's profile. Instead of `team` being a special collection separate from the rest of the site, it's now right there with `site.posts` and `site.pages`. This was actually the only, build-failing, page breaking failure that came out of the Jekyll upgrade (not bad considering they don't promise backward compatibility).

Apart from that, we got about a hojillion deprecation warnings on the site build (and continue to get some) because of our plugins needed updating. There's still one plugin, `jekyll-archives` that needs to be updated in order to squelch some lingering errors in the build.

If you build the site now (or look at the Travis output for this PR) you'll see this:
```
Deprecation: Collection#each should be called on the #docs array directly.
                    Called by /Users/gregoryboone/.rvm/gems/ruby-2.2.2@18f.gsa.gov/bundler/gems/jekyll-archives-659f024086ca/lib/jekyll-archives.rb:134:in `post_attr_hash'.
```

You'll see it roughly 143 times. It's because of a bug in the jekyll-archives plugin that is basically the same as the one we fixed on the team page. There is [a pull request in to that project](https://github.com/jekyll/jekyll-archives/pull/53) to address this.

Technically we could merge this now and just be cool with the deprecation notices. The downside is that people who build the site often (like @awfrancisco @elainekamlley or me) will see them _constantly._ I vote we hold this open until the fix is in for the archives, update the gem, and then test and merge.


konklone:
It seems like the PR you're waiting on is both active and reaching a conclusion, so :+1: on waiting for now.

gboone:
Indeed and the PR was just merged. Now I need to do some template updating. Will revisit this on Monday. Luckily, nothing urgent in here.

konklone:
jumanji5

gboone:
Alright! Let's get this merged! @18F/blog you will likely need to run `./go update_gems` the next time you build the site. 

More advantages in here: 
* Build times are now around 8 seconds
* `./go serve` now automatically rebuilds again (seems like auto-generation slowdowns are less of a thing in 3.0.0)
* `./go serve` and `build` commands now build incrementally, meaning Jekyll only re-generates the pages it needs to instead of rebuilding the entire site every time
* `reset` command in the go script will run `jekyll clean` then `jekyll build`. This will clear out your existing build and the build cache before rebuilding. This should be useful if `serve` or `build` run into problems and could be useful to run if rebuilds start running long.

elainekamlley:
Whoo :tada: 

Tested this and everything seems good to go. Updated the gems and built the site in about 8 secs. Confirms @gboone's notes. Merging!

PR #1285, opened 2015-11-03
Adds instructions for editing the authors.yml file

gboone:
cc: @18F/blog


awfrancisco:
Maybe add a section for outside contributors? Specifically that you can also change the URL linked to from their name? Or does that not really happen enough to include here?

gboone:
Done. Good idea.

PR #1288, opened 2015-11-03
Fix HTML errors

fulldecent:


gboone:
Thanks for catching these errors, @fulldecent!

PR #1301, opened 2015-11-09
DONT MERGE A file of only used CSS

msecret:
Just to show Elaine. We should take the used CSS file and do a diff
to try and find things we can remove.

Steps:
- `npm install`
- `(sudo) npm install -g grunt-cli`
- `grunt`
- New css file "tidy.css" in ./


gboone:
This is obsolete now that we're using the USWDS on the new site.

PR #1309, opened 2015-11-17
Replaces dots with dashes

gboone:
Dots in markdown filenames were generated as html files with dashes, e.g., `_team/lane.becker.md` generated as `_team/lane-becker.html`. The urls on `/team` were written with the `.` in the filename, e.g., the url for "Lane Becker" went to `/team/lane.becker/`. This caused the urls to 404 even though the pages were generated. I'm not sure why the tests didn't catch this, but I'll endeavor to figure that out.


mbland:
Odd... So the newer usernames are intended to reflect our GSA email addresses.

gboone:
I think that's a good idea, but we'll need to build in some kind of automatic character replacement in the future to sub dots for dashes, or change the dots to dashes on the team-api. I'm kind of agnostic on which is the right way to go at this point. 

[According to our content guide](https://pages.18f.gov/content-guide/urls-and-filenames/) we should be separating words with dashes in URLs and I want to make sure 18f.gsa.gov is following that convention.

mbland:
That's a good point. We can totally adopt that in the Team API as well. Do you want to file an issue in 18F/team-api.18f.gov?

gboone:
@mbland I'll definitely do that.

PR #1312, opened 2015-11-17
Publishes: Choose design over architecture

awfrancisco:


gboone:
This PR also adds new team member profiles (but not their images yet) and patches a previous uncaught accessibility problem in an older post. I am in the process of adding a fix to our deployment script that will address the "uncaught" part of the accessibility problem in the future.

PR #1315, opened 2015-11-17
Hides the unlabeled field from the DOM, addresses #a11y issues

gboone:
We don't need to show that, and it's probably better that we don't anyway.

@adelevie this is a fix for the accessibility issue you reported on #blog this afternoon. It turns out we were not hiding a field that didn't need any user input anyway. This fixes the `input` to make it `type="hidden"` instead of `type="text"`. cc: @nickbristow @18F/blog


gboone:
Also confirmed with @awfrancisco that this does not break the newsletter.

PR #1314, opened 2015-11-17
GSA.gov refresh post and new author

awfrancisco:
@18F/blog

PR #1308, opened 2015-11-17
New batch of photos for website bios

jenniferthibault:
### Changes

New photos for: 

@andrewmaier @mogul @brittag @OpenGlobe @xtine @fureigh @jessieay @mikeehsu @mzia @nicoleslaw @StephaniePlageman @timothybjones 

Small size edits to photos for:
 
@hestherchang , @jeremiak , and @annahsebok  

### Review:
@gboone

PR #1316, opened 2015-11-17
Publishes GSA.gov refresh cross-post

gboone:
Deploys #1314 #1313 and #1308 to production.


fureigh:
> removes the duplicate fureigh

:) Thanks, you beat me to it. Safe to say this is the first time there have been two Fureighs.

PR #1317, opened 2015-11-18
Jekyll upgrade, Gemfile housekeeping

### LINKED ISSUE #1286: Evaluate required gems in Gemfile
There are several gems we might not need anymore in order to run the build. `bourbon`, for example, is likely not needed for building the whole site. And `rouge` _might_ be installed as a dependency of Jekyll, making our inclusion of it redundant.

We should move as many dependencies out of the Gemfile as possible to make the stack a bit simpler. See also: https://gemnasium.com/18F/18f.gsa.gov

gboone:
Closes #1286 and upgrades Jekyll to 3.0.1 which contains some important [bug fixes](http://jekyllrb.com/news/2015/11/17/jekyll-3-0-1-released/).

This also removes the `bourbon`, `hash-joiner`, and `safe-yml` gems, and classifies `pry` as a development-only gem. We don't need pry running in production and we don't need the other three to build the site. The `bourbon` gem is only required to install the files required for the [bourbon framework](), so it has no use in the day-to-day running of the site. The `hash-joiner` and `safe-yml` gems were part of plugins and other scripts we have since eliminated.


Gemfile
```diff
@@ -16,9 +15,7 @@ group :jekyll_plugins do
...
 
-
-gem 'hash-joiner'
-gem 'safe_yaml'
-
 gem 'jemoji'
-gem 'pry'
+group :development do
+  gem 'pry'
+end
```
afeld: Are you doing much debugging of the plugins? If not, I don't think you need to include pry at all.

Gemfile
```diff
@@ -16,9 +15,7 @@ group :jekyll_plugins do
   gem 'jekyll_frontmatter_tests'
 end
 
-
-gem 'hash-joiner'
-gem 'safe_yaml'
```
afeld: Someone can't tell if these gems were needed or not from a glance, but seeing as [the site builds](https://travis-ci.org/18F/18f.gsa.gov/builds/91811785), doesn't seem that they are missed!

afeld:
Just the couple of comments – otherwise, looks good! :+1:

gboone:
@18F/blog this is ready to merge

PR #1331, opened 2015-11-21
Fix inconsistent reference to overlap time

waldoj:
I'm assuming that 4 hours is the correct number, and not 5 hours, because the time range "1 p.m. to 5 p.m. ET." is specified.


melodykramer:
You are right! It should be four. Thanks for fixing @waldoj!

PR #1335, opened 2015-11-25
Use Jekyll Feed

benbalter:
From the [original post](https://github.com/blog/2053-easier-feeds-for-github-pages);

> While it's always been possible to generate feeds using hand-written XML, the Jekyll Feed plugin includes a battle-tested template that handles things like encoding issues and relative URLs, allowing you to concentrate on what matters: your content.

Compare https://github.com/jekyll/jekyll-feed/blob/master/lib/feed.xml with the feed template in this repo (which I used to use before creating jekyll-feed). The idea is that this is a common problem that should be solved once and tested in a bunch of edge cases ([at least 500 sites, in this case](https://github.com/search?utf8=%E2%9C%93&q=jekyll-feed+filename%3A_config.yml&type=Code&ref=searchresults)).

The output should be the same for end users, with better cross-reader compatibility.


konklone:
@benbalter Mind adding an updated `Gemfile.lock` to the PR too? It's causing [the build to fail](https://travis-ci.org/18F/18f.gsa.gov/builds/93264385).

benbalter:
> Mind adding an updated Gemfile.lock to the PR too? It's causing the build to fail.

Done via 22c4ed1.

gboone:
Thanks @benbalter! I've been meaning to test this out. You have just forced my hand. :smile: _reviewing_

PR #1336, opened 2015-11-25
Add jekyll-seo-tag

benbalter:
This pull requests adds the Jekyll SEO Tag plugin.

You can [read more about specifically what the plugin does](https://github.com/benbalter/jekyll-seo-tag#what-it-does).

Similar to https://github.com/18F/18f.gsa.gov/pull/1335 it replaces the bespoke template in this repository with [a shared template](https://github.com/benbalter/jekyll-seo-tag/blob/master/lib/template.html).

This should support everything currently supported, plus next/previous URLs and JSON-LD post summary and site metadata. 

If there's anything the plugin doesn't do that you'd want it do to, I'd be glad to add it.

You can see example output [on my personal site](http://ben.balter.com) (output is explicitly commented).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :jekyll_plugins do
   gem 'jekyll-paginate'
   gem 'jekyll-redirect-from', '~> 0.9'
   gem 'jekyll_frontmatter_tests'
+  gem 'jekyll-seo-tag'
 end
 
 gem 'jemoji'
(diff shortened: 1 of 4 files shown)


konklone:
We'll need to review this one carefully, as we've put a lot of time into optimizing our various OG/SEO tags.

gboone:
Thanks again, @benbalter! 

Digging in: 
* We have a merge conflict now (gonna hazard a guess and say it's in the Gemfile.lock because _it always is_.)
* From the looks of the docs it seems like our open graph stuff won't break, but I'll definitely test it before merging.

gboone:
We'll make sure this gets into the new redesign.

benbalter:
> We'll make sure this gets into the new redesign.

@gboone :metal: where's that happening?

gboone:
Stay tuned here. :)

PR #1332, opened 2015-11-23
Update 2015-10-15-best-practices-for-distributed-teams.md

melodykramer:
Fixed subtraction error.

PR #1337, opened 2015-11-27
Deploy jekyll upgrade, typos to production

### LINKED ISSUE #1333: URLs recently became case sensitive
I have a link to https://18f.gsa.gov/2015/02/25/We-Are-Hiring/, which recently started 404ing. 

https://18f.gsa.gov/2015/02/25/we-are-hiring/ continues to work.

I suspect URLs *shouldn't* be case sensitive (although they can be), and that the change wasn't intentional.

gboone:
Quite a few bug fixes in here, all of which have already been tested and vetted on Staging.

* Fixes the broken urls caused by jekyll 3.0.0 by upgrading to 3.0.1 (see #1333 for the back story). We should probably lowercase those URLs at some point and setup the redirect for them so that they better conform to the [18F/content-guide](https://github.com/18F/content-guide).
* Fixes the time range in our distributed teams post

Closes #1333 and deploys #1317 #1332 and #1331 to production. Thanks to @benbalter @parkr @konklone @rebornix for their Thanksgiving Day help on this.


parkr:
Excellent!

gboone:
And if we wait until it's merged we could throw #1338 in here, too.

PR #1349, opened 2015-12-02
Resize Victor's image

jeremiak:
Victor's image was 300x300 which was making it bleed over onto his blog posts. Now its 250x250, which seems to be the size of everybody else's image.

Screenshot of bug:
screen shot 2015-12-02 at 9 23 26 am


gboone:
Good catch, @jeremiak

PR #1359, opened 2015-12-04
Update will.md

wslack:
Bio update! @melodykramer can you check? (unsure who to ask)


gboone:
Looks good from here.

PR #1365, opened 2015-12-07
Avoid framing the benefits of openness as one of "surveillance"

konklone:
I totally get the idea here, and I do talk about the helpful benefits of our staff feeling vaguely watched as they do their work. As a government team, I think it'd be best to avoid references to CCTV cameras and surveillance as explicit analogies for this effect, though.

### DIFF
--- a/_posts/2015-12-07-what-exactly-do-we-even-do-all-day.md
+++ b/_posts/2015-12-07-what-exactly-do-we-even-do-all-day.md
@@ -86,12 +86,7 @@ There are five major reasons why you should be this transparent too:
 
 ## 1. Working in the open drives higher quality results.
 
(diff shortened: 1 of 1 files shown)


gboone:
I agree with the impulse to avoid surveillance analogies, I think in this case it helps as an example of something specific in the "studies." @zacharycohn any thoughts on a different example we could use in place of surveillance here?

zacharycohn:
Didn't we already sort of miss the boat on this?

konklone:
I would still recommend changing it.

awfrancisco:
I think it's still fine to make this update if it's important. @zacharycohn Do you have any other examples we could use?

zacharycohn:
We can edit the first line to " Simply knowing someone could be seeing your
work has a demonstrable effect on behavior."

PR #1372, opened 2015-12-09
Rewrites cloud.gov paragraph on open source post

konklone:
The paragraph here about cloud.gov was confusing to me -- I wasn't sure what it meant after I read it. This is a second pass that tries to make it more clear, and cuts the line about what cloud.gov saves for the taxpayer to something more terse.


cmc333333:
:+1:

PR #1384, opened 2015-12-15
Adds a bio for Jessie Y.

gboone:
This is a bio for @jessieay based on her responses in the Google Form.

Attn: @awfrancisco @18F/blog and @18F/content


jessieay:
nice! this is not live on the site yet - anything we need to do to get it there?

awfrancisco:
@jessieay It will go live in the next PR from staging to production. That usually happens about once a day.

PR #1386, opened 2015-12-15
Adds a bio for Mossadeq Z

gboone:
cc: @18F/blog @18F/content


gboone:
This is ready to :shipit:

PR #1395, opened 2015-12-17
Publishes the Agile BPA launch post

awfrancisco:
Adds agile bpa tags to old posts


gboone:
This will also send the latest Agile BPA post to production.

PR #1416, opened 2015-12-28
Brandon Kirby's bio!

gboone:
@brandocalrissian's bio for 18f.gsa.gov, cc: @18F/blog @18F/content


gboone:
I think this is ready to merge

PR #1418, opened 2015-12-28
Adds a bio for Nicole

gboone:
@nicoleslaw cc: @18F/blog @18F/content


gboone:
I think this is ready to merge.

PR #1420, opened 2015-12-28
Removes a few team members who have moved on, updates Ric's bio

gboone:
Ric is no longer on leave, and we recently had to say goodbye to a couple people. This updates the team.


elainekamlley:
:wave:  Steph, @jackiekazil, and @OriHoffer! See you all around!

PR #1429, opened 2015-12-30
Adds a filter for youtube embeds

gboone:
This will make it easier for editing in federalist's editor eventually. In the meantime it will make sure all youtube embeds are properly scaled at smaller screen sizes. cc: @18F/blog

### DIFF
--- a/_plugins/embeds.rb
+++ b/_plugins/embeds.rb
@@ -1,0 +1,11 @@
+module Jekyll
+  module EmbedFilter
+    # Pass a url like {{ "https://youtube-nocookie.com/blahblahblah/" | embed }}
+    # It will spit out a standardized embed
+    def embed(input)
+      "<div class='embed-container'><iframe src='#{input}' width='560' height='315' frameborder='0' allowfullscreen></iframe></div>"
+    end
+  end
+end
+
+Liquid::Template.register_filter(Jekyll::EmbedFilter)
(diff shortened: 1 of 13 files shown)


assets/_sass/components/_embeds.scss
```diff
@@ -1,6 +1,6 @@
 .embed-container {
   position: relative;
-  padding-bottom: 100%;
+  padding-bottom: 56.25%;
```
rogeruiz: Would there also need to be a `padding-top: 25px;` to allow for the YouTube chrome on the embedded player?

_plugins/embeds.rb
```diff
@@ -0,0 +1,11 @@
+module Jekyll
+  module EmbedFilter
+    # Pass a url like {{ "https://youtube-nocookie.com/blahblahblah/" | embed }}
+    # It will spit out a standardized embed
+    def embed(input)
+      "<div class='embed-container'><iframe src='#{input}' width='560' height='315' frameborder='0' allowfullscreen></iframe></div>"
```
gboone: OK ignoring this because it's an HTML string and the length doesn't impede readability.

rogeruiz:
Looks good to merge in @gboone :+1:

assets/_sass/components/_embeds.scss
gboone: Good idea!

gboone:
thanks @rogeruiz!

Issue #292, opened 2014-10-05
Implement OCSP stapling


konklone:
[OCSP stapling](https://en.wikipedia.org/wiki/OCSP_stapling) is a method for checking whether the SSL certificate for a website has been revoked at the very same time as you connect to the website. The revocation data is "stapled" to the SSL response at connection time.

Without OCSP stapling, browsers which do revocation checking (e.g. Firefox, but not Chrome) have to hit a web API that SSL certificate authorities provide, that offer a list of revoked certificates. Doing this is slow, has poorer security properties, and is a privacy leak (it shares browsing data with CAs).

It's not yet enabled in our nginx configuration, as I have a [little more research](https://github.com/18F/DevOps/tree/master/nginx#ocsp-stapling) to do -- especially in figuring out the instrumentation to test whether OCSP stapling is actually working.

konklone:
A note that SSLMate has an excellent blog post on OCSP stapling implementation in nginx and Apache:

https://sslmate.com/blog/post/ocsp_stapling_in_apache_and_nginx

konklone:
This is an upstream request of Federalist, but I think CloudFront supports OCSP stapling, so this may be solved for now.

wslack:
@konklone should we have an issue in the federaist repo?

konklone:
@wslack Yeah, it looks like it. SSL Labs says no OCSP stapling for 18f.gsa.gov: https://www.ssllabs.com/ssltest/analyze.html?d=18f.gsa.gov&s=52.84.213.151

PR #1430, opened 2015-12-30
[WIP] Federalist prep

gboone:
This pull request preps 18f.gsa.gov to use the Federalist platform. The short term goal is to use the previewing feature for, among other things, blog post drafting. The long term goal is to fully host the site on the platform. Outstanding issues blocking that include:

- [x] Integration of 18F/Dashboard at `{{ site.baseurl }}/dashboard/`or dashboard.18f.gov
- [x] Ensuring there are no `baseurl`-less URLs, images, links, scripts, etc. kicking around.
- [x] Clearing out all html files that might be edited as content (convert them to markdown)
- [x] Understanding any implications this move might have for our ssl configuration, compliance, or other devops concerns. AFAIK there are none but this, especially compliance, is always a concern.
- [x] Figure out how to get URLs to respond with or without the trailing slash (`{{ site.baseurl }}/team` vs `{{ site.baseurl }}/team/`)

Preview branch: https://federalist.18f.gov/preview/18F/18f.gsa.gov/federalist-prep/


gboone:
Aaand Travis failed showing all our internal links as broken. Interesting. Looks like I have a todo list for tomorrow.

gboone:
@jeremiak @dhcole I'm curious to hear your thoughts on something @elainekamlley and I were just talking about w/r/t the Dashboard. Right now the site and the Dashboard are wholly independent. The Dashboard is a Jekyll site with [a separate NGINX location](https://github.com/18F/18f.gsa.gov/blob/staging/deploy/18f-site.conf#L43-L46) configured to serve it at `/dashboard/` for both staging and production. I'm thinking there are two options for it if we move to Federalist:

1. Install the it as a submodule or otherwise build it into the site's deployment routine in Federalist. I'm thinking something similar to how to can specify a `post_build` in a Travis config. (Don't even know if there are plans to support that kind of thing.)
1. Move the Dashboard to a different server like dashboard.18f.gov and redirect all the /dashboard URLs. I think there's a UX advantage here that I want to explore a little more. We've had a many people confused by the context change when moving from other pages on 18f.gsa.gov to /dashboard, a change of URL could be enough of a different to indicate "you're somewhere else now."

dhcole:
Running them as separate sites on separate subdomains seems the most clean approach to me. I think I'd go that to keep the pieces logically separated.

If you prefer the dashboard in a nested subdirectory, as it is now, we would do that with a proxy or CDN.

For other Federalist sites, we're using CloudFront as a CDN for SSL termination and routing. For this site, I think we'd need to keep the nginx server in front but change its configuration to simply proxy the Federalist S3 origin rather than serve files. This would preserve the custom headers on the site that we're not able to set with CloudFront:

- strict-transport-security:max-age=31536000; includeSubDomains; preload
- x-content-type-options:nosniff
- x-frame-options:DENY
- x-government-innovation:Disrupted
- x-xss-protection: 1; mode=block

Ideally we'd have support for these in CloudFront, but we haven't been able to get that from AWS yet.

Either way, we can easily have nginx proxy to two Federalist sites if you'd like to keep the subdirectory.

When you're ready to launch, I'm happy to help with nginx if you'd like. I've already set up the configuration for this in a Cloud.gov-ready application: https://github.com/dhcole/s4

dlapiduz:
Huge :+1: on moving the site into federalist, please let me know if I can help on any way.

Some of the headers should come from cloud.gov built in so if you do the S4 proxy then you should be set.

As for the dashboard I agree with @dhcole on having it as a separate domain and site. There is no need to have it as a subfolder.

If it is intended to be as a subpath then why not have it all in the same repo?

gboone:
Very interesting information. @dlapiduz or @dhcole do either of you have an example of how that S3 proxy works you could share? I only sort of understand how it would work.

dhcole:
@gboone: using https://github.com/dhcole/s4 as an example, you just need to set two env vars and deploy to Cloud.gov.

```
BASE_DOMAIN: 18f.gsa.gov
PROXY_ROOT: http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/18F/18f.gsa.gov
```

Where `BASE_DOMAIN` is your custom domain and `PROXY_ROOT` is the Federalist origin path (available under site > settings) Note: it should not have a trailing slash for this application.

The relevant configuration is here: https://github.com/dhcole/s4/blob/master/nginx.conf#L29-L45

gboone:
What do you think the timeline would be like to move this over? End of the week? End of the month? End of the _DAY_?

dlapiduz:
@gboone if you are going to use cloud.gov for the proxy we need to set up an ELB with the SSL cert. It doesn't take more than a day but it is not fully automated yet so it may take a couple hours or so.

gboone:
I'm also leaning toward deploying dashboard at dashboard.18f.gov. I've got a PR coming in today that _should_ allow us to serve it over cloud.gov. I'm not sure it needs to be on Federalist right now, but perhaps in the future hosting it there would be valuable.

dlapiduz:
@gboone didn't see this comment... we can move the dashboard to cloud.gov without federalist, it would make things a bit simpler...

konklone:
The S4 proxy is a great solution for the HTTP headers, and the ELB can host the Let's Encrypt cert we're using.

With an ELB in front, we'll lose control over ciphers and our use of HTTP/2, right? Is there any way to support HTTP/2 on a cloud.gov-hosted app at the moment?

One thought - what happens if an ELB in front of Cloud.gov is configured for straight TCP passthrough? Can Cloud Foundry still successfully proxy that ciphertext into the S4 nginx proxy for TLS termination there?

dlapiduz:
@konklone unfortunately we can't do HTTP2 at the moment. Basically even if you did TCP out of the ELB there are 2 more layers of routing and proxying inside CF until it hits your app.
One of those expects HTTP traffic so we can't do HTTP2 there, yet.

I am looking into implementing a solution that would help us do HTTP2 with automagic LE support but I haven't even started it.

gboone:
Given all that, I think we'd be very much interested in supporting HTTP2 work on cloud.gov as part of the soon-to-begin redesign work.

Is it possible to renew the LE cert every three months on cloud.gov or should we plan to go back to our non-LE cert temporarily?

dlapiduz:
@gboone so HTTP2 is going to happen but 18f site wouldn't be a priority any different than the whole platform. It is going to be a while until we have this done.

For LE it is easier to do this kind of stuff in a single server but I believe we can figure out a way to do it with a combination of nginx rerouting and S3.

dhcole:
Are HTTP2 and LE blockers to migrating? I think we're ready to do this otherwise.

konklone:
One more question - @dlapiduz , how challenging would it be to have an EC2 instance do the termination and proxying to Cloud Foundry, instead of an ELB? The instance's role would be quite simple, and wouldn't need to have any state about the site it's proxying for, except for the hostname and certificate. This would get 18f.gsa.gov the deployment benefits and featureset of Cloud.gov and Federalist, while still leaving the site control over its termination. As I understand it, the terminating ELBs for Cloud.gov are managed off-band from the rest of Cloud.gov deployment anyway.

gboone:
Neither are blockers, no. I'd like to keep our LE cert if we can, but if not, or if it's going to be difficult to renew it in a month (or so) I'm OK with forging ahead.

dlapiduz:
@konklone it wouldn't be challenging... you wouldn't need cloud.gov at all. You can just proxy to S3.
But I think the whole point of what we are trying to do is to get away from single EC2 instances and move to more scalable (at least in terms of people process) solutions.

I would try to focus our effort in making sure that we advance the platforms for everyone rather than trying to find hacks just for the 18F site.

gboone:
> I would try to focus our effort in making sure that we advance the platforms for everyone rather than trying to find hacks just for the 18F site.

How can a future dev team for this site help make that happen? Is it a we-don't-have-enough-people problem and we could commit someone from the team to taking on cloud.gov tasks?

dlapiduz:
@gboone for a long time we had a very small team and we are now catching up. The team has grown exponentially in the back end and hopefully we will soon be able to tackle things like this.

That said, if you have people interested in working on stuff like this we can figure something out.

gboone:
That's great. Once we get up and running I'll let you know what that might look like.

@dhcole I'm working on converting the remaining html files in _posts to markdown but I think we're ready to get this off the ground. Since I'm OOO next week maybe we should hold the full cutover until I get back online. Just so I can be on hand for the blog team if they need post-transition support. I think this is good to merge, though.

gboone:
OK. I think this is ready to merge. Before making the full cutover we need to:
- [ ] Get Dashboard squared away at dashboard.18f.gov
- [ ] Redirect all current /dashboard pages to their corresponding URL on dashboard.18f.gov
- [ ] Fix this trailing slash thing. All internal links include a trailing slash but it'd be great if someone could type https://18f.gsa.gov/blog into their blog and not get a weird file download instead of the blog.

dlapiduz:
@sharms is working on helping the dashboard team move to dashboard.18f.gov btw

gboone:
Awesome. @sharms look forward to working with you on that :) Work has begun here: https://github.com/18F/dashboard/pull/339 happy to discuss our hopes and dreams, fears and anxieties, there.

gboone:
OK, the trailing slash thing I _think_ is a problem only with the preview URLs. If you go to the S3 bucket directly, the downloading problem doesn't happen. Hat tip to @shawnbot for helping figure this out. If you navigate directly to `.../blog` on the S3 bucket it properly redirects to `.../blog/`.

konklone:
As a note, another thing we should bear in mind about moving to ELBs is we will not be able to do [OCSP stapling](https://github.com/18F/18f.gsa.gov/issues/292). 

The lack of stapling is what causes the unencrypted OCSP request to ocsp.x1-int.letsencrypt.org here: http://www.webpagetest.org/result/160113_2S_1207/1/details/ It's also one of the third party services I [originally proposed documenting](https://github.com/18F/18f.gsa.gov/pull/399).

Stapling for 18f.gsa.gov only stalled because, at the time we were talking about it, nginx had done a very confusing and incomplete job of implementing it, but that has been addressed in recent versions of nginx and there are [clear best practices now](https://sslmate.com/blog/post/ocsp_stapling_in_apache_and_nginx).

It's up to Amazon to enable OCSP stapling for ELBs, and I haven't heard anything from them about doing this. They do have it enabled for CloudFront.

gboone:
We can safely merge this without affecting the existing hosting environment. That will at least give us the advantage of preview branches and potentially the editor while we figure out how to move forward on these remaining issues.

_layouts/tag_index.html
```diff
@@ -10,7 +10,7 @@
     {% for post in page.posts %}
         <article class="tag-index">
           <div class="blog-title">
-            <h1 itemprop="headline" class="h4"><a class="link-alt" href="{{ post.url }}">{{ post.title }}</a></h1>
+            <h1 itemprop="headline" class="h4"><a class="link-alt" a href="{{ site.baseurl }}{{ post.url }}">{{ post.title }}</a></h1>
```
dhcole: Extraneous  `a` attribute in the `anchor` tag

_layouts/tag_index.html
```diff
@@ -37,7 +37,7 @@ <h1 itemprop="headline" class="h4"><a class="link-alt" href="{{ post.url }}">{{
           </div>
 
           <div class="blog-continue">
-            <p><a href="{{ post.url }}">Continue reading this post...</a></p>
+            <p><a a href="{{ site.baseurl }}{{ post.url }}">Continue reading this post...</a></p>
```
dhcole: Extraneous `a` attribute in the `anchor` tag

dhcole:
Nice work @gboone. Looks great to me! :ship: :rocket: :shipit:

dhcole:
@gboone to close out the rest of this conversation, the hosting piece, I set up the proxy. It's available here: https://18f-site.18f.gov/

It's running as the app `18f-site` in the `federalist` org, `prod` space on cloud.gov. It's proxy with 2 instances for redundancy, so it should be ready to go.

To switch the main domain to this proxy, there are two steps:

1. Change the custom domain in the Federalist site settings to `https://18f.gsa.gov` and change the `BASE_DOMAIN` ENV var to `18f.gsa.gov` for the `18f-site` cloud.gov app. @jeremiak can help with this. 
2. You'd need to add a route for `18f.gsa.gov` to the cloud.gov app and set up an ELB to terminate the SSL cert, and adjust the DNS settings to point to that ELB. @dlapiduz should be able to help with that — it's all [standard procedure for custom domain](https://docs.cloud.gov/apps/custom-domains/) cloud.gov applications.

Also the proxy repo is now here: https://github.com/18F/cg-s3-proxy and is deployed with the following ENV vars:

```
BASE_DOMAIN: 18f-site.18f.gov
PROXY_ROOT: http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/18F/18f.gsa.gov
```

dhcole:
Also, at that point, you may want to switch your default branch in Federalist from `staging` to `production`.

PR #1447, opened 2016-01-07
Adds a redirect from the old date to the correct url

gboone:
This should redirect any requests for the post with the old date to the correct place.


gboone:
@dhcole Will this solution continue to work once the site is deployed on federalist or should we use a fully qualified URL for the redirect?

dhcole:
Yes, it will work as long as the gem stays included in the gemfile: https://github.com/18F/18f.gsa.gov/blob/staging/Gemfile#L13

gboone:
Awesome :+1:

PR #1464, opened 2016-01-19
New bios

gboone:
This adds a bunch of new markdown files for people with missing bios. cc: @jeanninehunter you can add their bios to this PR if you'd like or wait for it to get merged.

It also fixes the script that imports them from the Team API to account for a bunch of `name` keys that changed. Apparently Team API didn't like the names we were using :wink:.  The script will now look for `deprecated_name` and use that first. If there's no `deprecated_name` it uses `name`. The script isn't perfect. If someone's `name` is changed without adding a `deprecated_name` of if someone's `deprecated_name` is spelled incorrectly, for example, we'll get duplicates. See also: 18F/team-api.18f.gov#173.


gboone:
OK, the hound should be satisfied with the changes in the last few pull requests and the build should start passing with 6cbd960

gboone:
This is ready to go

PR #1467, opened 2016-01-21
Deployment: Migrate hookshot to githooked

ScottONeal:
The author for [hookshot](https://github.com/coreh/hookshot) has been inactive and not accepted further contributions. I decided to fork hookshot and continue development of the module, the fork includes BDD tests, more annotations, and better integration of nodes `EventEmitter`.

githooked repo: https://github.com/ScottONeal/githooked
githooked npm package: https://www.npmjs.com/package/githooked

Part of the PR includes updates to `deploy/githooked.js` (formerly `hookshot.js`), which now uses commander for CLI parsing. I made a similar to PR to 18F/pulse: https://github.com/18F/pulse/pull/212

This PR also includes a small `package.json` in `deploy/`, so that proper dependency integration can be managed, this will mitigate node modules introducing breaking changes (semver being respected).

Mock deployment tests using githooked worked in development without any api changes.


deploy/fabfile.py
```diff
@@ -6,15 +6,15 @@
...
-  forever restart deploy/hookshot.js -p 4000 -b production -c "cd $HOME/production/current && git pull && bundle exec jekyll build --config _config.yml,_deploy_config.yml >> $HOME/hookshot.log"
-  forever stop deploy/hookshot.js -p 4000 -b production -c "cd $HOME/production/current && git pull && bundle exec jekyll build --config _config.yml,_deploy_config.yml >> $HOME/hookshot.log"
+  forever start -l $HOME/githooked.log -a deploy/githooked.js -p 4000 -b production -c "cd $HOME/production/current && git pull && bundle exec jekyll build --config _config.yml,_deploy_config.yml >> $HOME/githooked.log"
+  forever restart deploy/githooked.js -p 4000 -b production -c "cd $HOME/production/current && git pull && bundle exec jekyll build --config _config.yml,_deploy_config.yml >> $HOME/githooked.log"
+  forever stop deploy/githooked.js -p 4000 -b production -c "cd $HOME/production/current && git pull && bundle exec jekyll build --config _config.yml,_deploy_config.yml >> $HOME/githooked.log"
```
gboone: For these and the staging hooks, would you mind switching the log file back to `hookshot.log`? We have a log rotator that I think watches for "hookshot.log" @mbland is that correct?

deploy/fabfile.py
```diff
@@ -31,7 +31,7 @@
 env.hosts = ["18f-site"]
 
 home = "/home/site"
-log = "%s/hookshot.log" % home
+log = "%s/githooked.log" % home
```
gboone: here, too, githooked maybe should remain hookshot (cc: @mbland)

gboone:
@ScottONeal this is great, thanks for contributing back to this project. I noticed that hookshot had gone stale and was just wondering about this myself. @konklone you reviewed the other PR's he mentioned, do you see anything in here that would block a merge? I left a couple line comments but other than those it looks good to me.

Thanks again, @ScottONeal!

ScottONeal:
Cool, I'll wait for advisement from @mbland before making log file changes.

konklone:
@gboone This also looks good to me. We merged in https://github.com/18F/pulse/pull/212, as it looked good to us, though we've also been "unhooked" since some major refactoring, so we're not actually exercising the webhook code in production. That also was just adding the use of `commander`, not transitioning to `githooked`.

That all said, this PR looks great, and I'm really glad @ScottONeal took charge of the situation and breathed some life into a helpful tool!

gboone:
OK, only outstanding question is the log files. @mbland will the log rotation we have setup rotate all `.log` files or only `hookshot.log`. If it's all log files we can merge this.

mbland:
We're not rotating those logs, but they're only about O(120K). I'd say go for it.

@ScottONeal I've got a couple other services that I could update to this from 18F/hookshot, which I hacked to pass options through to the JSON parser middleware.

ScottONeal:
@mbland Okay, a few days ago, I added the json limit param as a static value to githooked, since that seamed to be more of a bug of the module.

If you have other body-parser json params that fall under a similar scope, being more of a bug than a feature, than I'll do some testing and add it as a static param. Otherwise, I'll come up with an elegant way of passing express middleware configuration.

On another note, have you looked at using [PM2](https://www.npmjs.com/package/pm2) instead of forever? That's what we use at uscensus, seems to be a bit more robust and there is also a logrotate module, if needed.

konklone:
PM2 looks [a lot more active](https://github.com/Unitech/pm2/graphs/contributors), and it's probably a good idea.

gboone:
Merge this now and figure out switching to PM2 next?

PR #1476, opened 2016-01-27
Edit how to use github, SSH and Desktop App

michaelccata:
See #18f-site for details - adding copy to reflect SSH key generation through GitHub desktop log-in. 

Running Laptop without opening/registering GH desktop will result in failed cloning (and presumably other operations) due to a lack of local SSH key. Registering the account and validating with 2FA on desktop application will auto-generate an SSH.


gboone:
This looks great, thanks for helping keep this post up to date, @michaelccata!

PR #1478, opened 2016-01-28
Fix link to the Data.gov page for .gov domains

konklone:
The data.gov permalink for .gov domains changed. This fixes it.

cc @hkdctol


gboone:
Thanks @konklone

PR #1480, opened 2016-01-28
redirect old URL

gboone:


gboone:
@andre this works on staging. You're good to merge.

PR #1512, opened 2016-02-19
Adds Dashboard to 18f.gsa.gov repo

gboone:
Migrating Dashboard fully into 18f.gsa.gov. Will allow us to fully deploy on Federalist and begin to sunset the [Dashboard repo](https://github.com/18f/dashboard).


gboone:
Assuming the tests pass in 6eb585a we should be good to go on this one. @18F/blog and @jeremiak let's chat sometime tomorrow or Friday about making the full switch. I'm going to create a `master` branch and we can start working from there as our main branch that automatically deploys to 18f.gsa.gov.

This will mean moving to a one pull request blog publishing system, so we'll have to change our workflow to make sure we're sharing a Federalist branch with authors as the staging server.

PR #1516, opened 2016-02-22
Publishes the software maintenance is an anti pattern post

awfrancisco:


ultrasaurus:
heard from @awfrancisco that this has been approved, and looks good to me, so I will merge :shipit:

PR #1520, opened 2016-02-29
add images for federal front door post

melodykramer:
image00.jpg alt text: An infographic illustrating that people expect personal information like medical records, family information, and financial records to be safe from ID thieves, advertisers and third parties.

image01.jpg alt text: A two-by-two grid showing how people think when confidence is plotted against resilience. When confidence and resilience are both high, people "will keep trying until a goal is achieved." When confidence is high but resilience is low, people may apply once but will give up if it doesn't go well. With resilience high and low confidence, people may try a few times but will doubt themselves. And when both resilience and confidence are low, they probably won't apply at all.

image02.jpg alt text: An infographic showing sources of pressure and types of emotions a person might feel after different interactions with the government.

image03.jpg alt text: A graphic showing how people might view sharing information. Some never want their information shared, and others are on a continuum.


awfrancisco:
Do we have to re run travis?

Issue #411, opened 2014-12-05
Create our own 18F privacy policy


polastre:
Instead of leaning on the more generic GSA privacy policy, create our own privacy policy.

This policy should cover all of the standard privacy policy requirements, including our sharing of data (as #399 would like to do).

Bonus points for writing a privacy policy in plain english using common language that is easily understandable.  Even more bonus points for writing a policy based on user needs research that identify and informs people about what they can **do** now that they know about our privacy policy.

PR #1523, opened 2016-03-01
add ffd post 1 github issue #411

melodykramer:


gboone:
No idea why the html-proofer tests worked yesterday but not today. Updating the gem did the trick, though, so whatever it was is fixed now.

PR #1562, opened 2016-03-11
Fix double period at end of sentence

waldoj:
It looks like there was a plan to add another citation, but that wasn't followed through on. I removed the link that's attached to no text, leaving only the ones already showing up (presumably the ones that passed editing muster).


gboone:
Thanks @waldoj!

PR #1591, opened 2016-03-22
updates the production deploy to pull from the `staging` branch

gboone:
This PR essentially deprecates the `production` branch. Since we're doing manual deploys to production and eventually moving the whole stack to 18F/federalist, that branch is redundant and only slows down our publication flow. Merging has no *immediate* effect but will allow us to skip the second pull request and just run the deployment script to rebuild production.

### DIFF
--- a/go
+++ b/go
@@ -33,8 +33,7 @@ end
 
 def init
   begin
-    require 'bundler'
-  rescue LoadError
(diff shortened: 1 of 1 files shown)


gboone:
cc: @18F/comms

PR #1610, opened 2016-03-31
Fixes its

emileighoutlaw:
removes stray apostrophe


emileighoutlaw:
cc @gboone

gboone:
:+1: good catch @emileighoutlaw!

PR #1625, opened 2016-04-05
Added article end tag

digilou:
Minor markup update. Added </article> on line 63 within the "services" <section> because the <article> end tag was missing.


gboone:
Great fix, @digilou, thank you! I can't believe we missed this one. :blush:

PR #1631, opened 2016-04-08
Fix JSON formatting in ES post

konklone:
This updates [the Elasticsearch post](https://18f.gsa.gov/2016/04/08/how-we-get-high-availability-with-elasticsearch-and-ruby-on-rails/) by @pkarman to fix the JSON syntax highlighting.

It changes this:

image

Into this:

image


pkarman:
I'm fine with this.

alternately, could change the fencing label to be explicitly `ruby` instead of empty or `json`. afaik that would make the quotes not needed.

gboone:
Thanks @pkarman and @konklone

PR #1634, opened 2016-04-11
Removes the feedback widget

gboone:
Turns out it wasn't as reusable as we thought. @brittag can you verify I got the right one removed?


brittag:
Yes, this is right. Thank you!

PR #1641, opened 2016-04-15
link directly to the #devops-public chat room from the blog posts

afeld:
https://chat.18f.gov now supports linking directly to the channel! See https://github.com/18F/chat/pull/5 for more information.


awfrancisco:
That's awesome. We should probably look for all the other chat.18f.gov links and update them as well.

afeld:
Already done for everything I could find on GitHub 😉

awfrancisco:
💥

PR #1661, opened 2016-04-22
Update Clint Troxel

ctro:
- Changed role to Developer
- Simplified Description

FWIW updating this via PR is way easier than using the Google Form I was directed to use during onboarding.


gboone:
Thanks @ctro!

PR #1677, opened 2016-05-02
Swap carousel for static image

hbillings:
Also fixed a bunch of hound alerts.


awfrancisco:
Did this update the global nav too?

PR #1684, opened 2016-05-02
combine and minify styles

brendansudol:
__why__
we had 7 different stylesheets being loaded in the head, which creates additional
requests and negatively impact page load speeds.

__how__
we can use the built-in sass compiler to compile and minify all of these styles.

__before__
image

__after__
image


theresaanna:
LGTM! 👍

theresaanna:
One question, though -  are any documentation updates needed?

brendansudol:
thanks @theresaanna. i just looked through documentation -- i don't believe any updates need to be made and none of the scripts should be are affected

brendansudol:
thanks @theresaanna. i just looked through documentation -- i don't believe any updates need to be made and none of the scripts should be are affected

PR #1686, opened 2016-05-03
navbar short-term fixes

brendansudol:
This PR fixes a few UI bugs re: the current navigation.

I think we should consider further clean-up & refactoring of this component, but in the short-term wanted to address these small things.

Before:

image

image

image

After:

image

image


gboone:
Thanks @brendansudol!

PR #1691, opened 2016-05-05
fix layout re: page headers

brendansudol:
Fix some header spacing inconsistencies / crampedness

Before:

image

image

After:

image


brendansudol:
These hound errors are from the legacy sass file -- I did not change anything except for spacing; I think we can safely ignore :)

gboone:
Yes. Also, I'm going to turn Hound off until we're ready to stop ignoring it.

PR #1702, opened 2016-05-10
Restore the dashboard

gboone:
Looks like the Dashboard somehow didn't come with us when we moved to Federalist. This puts it right back where it belongs.


gboone:
ping

Issue #1705, opened 2016-05-11
Team Page giving a 404 error when calling from the page footer


neilmartis:
When I clicked on the page footer of the team page I am getting a 404 error
https://18f.gsa.gov/site/18F/18f.gsa.gov/team/
![18fteam](https://cloud.githubusercontent.com/assets/10228512/15181594/15db91e6-1756-11e6-9eef-eb3db0c24561.png)

brendansudol:
thank you!

fixed via https://github.com/18F/18f.gsa.gov/pull/1706

PR #1706, opened 2016-05-11
add trailing slash to footer links

brendansudol:
(this was causing 404s)


brendansudol:
This addresses https://github.com/18F/18f.gsa.gov/issues/1705

Issue #1704, opened 2016-05-11
header content does not fit in box when h1 is > 31 chars


jgrevich:
![screen shot 2016-05-10 at 8 32 43 pm](https://cloud.githubusercontent.com/assets/24955/15169272/64b97fc0-16ee-11e6-87e9-3363adba9c03.png)

The `h2` also wraps at > 52 and  > 93 chars but that's probably less of an issue. 

My wife found this and #1703. We should get her to help with QA :)

neilmartis:
@jgrevich Hi Justin This is happening on all pages I checked it yesterday as well

gboone:
This has been an issue since we rolled out these profiles. Luckily, few people have names and titles that long, but it's definitely something we should figure out when we revisit these pages. Thanks @jgrevich

gboone:
This will be addressed with the deprecation (and eventual re-thinking) of how we display our team. Thanks @jgrevich

coreycaitlin:
OBE

PR #1709, opened 2016-05-11
fix header content bug

brendansudol:
This PR addresses https://github.com/18F/18f.gsa.gov/issues/1704

Note: this is a short-term fix -- we need to go back and clean up and trim a lot of this CSS, but for now, this should fix the content fitting issue.

Before:
image

After:
image


gboone:
Thanks @brendansudol!

PR #1722, opened 2016-05-16
fix project page layout

brendansudol:
this should fix the current project page rendering / redirect issue


brendansudol:
this should fix the current project page rendering / redirect issue

PR #1738, opened 2016-05-20
[WIP] Spring 2016 IRL detailed agenda

gboone:
Don't merge until we've fully annotated the agenda, at least to the best of our ability.


_config.yml
```diff
@@ -93,6 +93,8 @@ collections:
   cards:
     output: true
     permalink: /card/:name/
+  irl:
```
afeld: Minor: To avoid cluttering the top-level directory, I would try to move this under some other directory.

afeld:
Faaaancy. Honestly, it seems overkill...a blog post with a Markdown table in it seems like it would achieve the same thing and be less of a special ❄️.

gboone:
Yeah, I went this direction to keep it easier to replicate for future IRLs, and to make the individual sessions easier to update if we need to change the summaries. Does feel a little much, though. Thanks for reviewing @afeld!

_config.yml
gboone: Yeah, with `output: false` I think it will not generate any of these (that's the hope anyway) into the site. We have one page at /irl/spring-2016/ for the whole thing and we loop through the collection docs to generate that page.

NoahKunin:
Is this still a thing?

gboone:
We'll revisit this in the new design and IA.

PR #1739, opened 2016-05-20
Change to to normal Twitter summary card for posts without images

maya:
This uses the normal sized twitter summary card unless an image is supplied in the post.

cc: @gboone


gboone:
Thanks @maya!

PR #1758, opened 2016-06-06
Link to "What is an API" post from the "Open Data Democratizes Innovation" post

toolness:
The [Open Data Democratizes Innovation](https://18f.gsa.gov/2016/06/02/open-data-democratizes-innovation/) blog post mentions what the acronym "API" stands for. I thought it might be nice to have that link to our post on [What Is An API?](https://18f.gsa.gov/2016/04/22/what-is-an-api/), in case the reader wants more details on that.


awfrancisco:
This was a cross post from non-18F authors, so I don't want to insert links without notifying them. But this is a good reminder that we should be linking other API references to that post.

PR #1761, opened 2016-06-07
Removes api.data.gov

awfrancisco:
cc @wslack


awfrancisco:
Please hold on merging

cmc333333:
We still manage api.data.gov

awfrancisco:
Yeah, this list isn't meant to be a comprehensive list of projects for each business unit, but rather examples of what we can sell to or do for agencies. The Product and Platforms list is platforms that agencies can take advantage of through IAAs, and it's my understanding (which might not be right) that we don't sell api.data.gov to other agencies, we simply manage it.

wslack:
This can be closed; it should have been on my list.

PR #1771, opened 2016-06-16
Use SVG image URLs

shawnbot:
This PR replaces the image src attributes of Emanual's renderings with the SVG URLs. [Every browser from IE9 up](http://caniuse.com/#feat=svg) supports SVG natively, and these files are much smaller than their JPEG equivalents. cc @melodykramer


awfrancisco:
We host all the images for our blog on our own site, so we should download and add the SVGs to our assets library. 

cc @gboone

gboone:
Pushed a commit to load the SVGs and cut one of the jpegs out. Looks good on the Travis side.

PR #1774, opened 2016-06-17
Fix OpenGraph metadata on DAP update post

konklone:
We left Markdown in the OpenGraph tags for the DAP post.


awfrancisco:
Thanks for the catch

PR #1778, opened 2016-06-21
Jekyll update

gboone:
Not sure why there are three versions of this commit but the result is that Jekyll gets updated to 3.1.6, the latest and greatest Jekyll there is.

We're upgrading from 3.1.2 to 3.1.6, here are the blog posts for the current and intervening versions.

3.1.6: https://jekyllrb.com/news/2016/05/19/jekyll-3-1-6-released/
3.1.5: https://jekyllrb.com/news/2016/05/18/jekyll-3-1-5-released/
3.1.4: https://jekyllrb.com/news/2016/05/18/jekyll-3-1-4-released/
3.1.3: https://jekyllrb.com/news/2016/04/19/jekyll-3-1-3-released/

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 ruby  '2.2.3'
 gem 'redcarpet', '3.2.3'
-gem 'jekyll', '3.1.2'
+gem 'jekyll', '3.1.6'
 gem 'html-proofer'
 
 group :jekyll_plugins do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -36,7 +36,7 @@ GEM
       yell (~> 2.0)
     htmlentities (4.3.4)
     i18n (0.7.0)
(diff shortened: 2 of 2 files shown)


monfresh:
I noticed that you've been updating the Gemfile each time you upgrade Jekyll. I'm not sure if that was on purpose, but I thought I'd share a tip that it's not necessary to update the Gemfile each time. You just need to decide once what versions you are willing to accept, and then run `bundle update [name of gem]`. If you feel confident that any 3.x version of jekyll will be compatible with your app, you can do this in your Gemfile:
```ruby
gem 'jekyll', '~> 3.1'
```
Then, when you want to upgrade jekyll, you would only run `bundle update jekyll`, which will automatically update `Gemfile.lock`, and you won't need to change the `Gemfile` as well.

gboone:
Thanks for the tip, @monfresh, we'll have to take a better look at versions and figure out how to manage it properly.

PR #1791, opened 2016-07-05
Gem updates: Jekyll plugins, redcarpet

gboone:
Upgrades the jekyll-feed plugin and redcarpet to incorporate bug fixes. The latest version of redcarpet no longer results in build failures, which is great! @monfresh should get eyes on this to make sure we're good and @18F/blog should take a look at some of the more complex blog posts to make sure the new version of redcarpet doesn't require any markdown changes. Compare this post: https://18f.gsa.gov/2016/06/24/5-lessons-in-object-oriented-design-from-sandi-metz/ with the preview generated for this branch.


monfresh:
Looks good to me, but it looks like HTML Proofer found some issues. Also, if you don't want those extra merge commits in your history, you can use `git rebase staging` onto your branch instead of merging.

gboone:
OK, the tests passed on my end so I'm expecting Travis to do the same. @monfresh would you squash and merge if Travis goes green on this build?

monfresh:
I can't merge until you rebase/merge staging in your branch because this repo requires that PRs be up to date with staging before merging. If staging changes frequently, I would recommend turning that setting off because you will end up having to rebase every time staging changes. Travis already notifies you of incompatibilities with staging via their `pr` build, which merges the PR with staging and runs tests against that.

gboone:
Updated from staging. Thanks @monfresh!

monfresh:
👍

PR #1795, opened 2016-07-06
When a micro-purchase doesn’t work out, we try to learn from it

gboone:
Indeed we do. Here's the post for publication tomorrow, pending final layout review by @mtorres253 and/or @vzvenyach


mtorres253:
Layout looks good to me @gboone

PR #1799, opened 2016-07-12
optimizes a whole bunch of images

gboone:
@maya @dogweather here are all the images optimized. Let's hold on merging until/unless we know they still look OK. We can always re-run imageoptim at a less aggressive setting (though we'll get less savings that way).


dogweather:
thanks, i'll take a look.

gboone:
An upcoming redesign has made this PR obsolete.

PR #1805, opened 2016-07-18
Cloud.gov is FedRAMP Ready

gboone:
Publishes "Cloud.gov is full steam ahead on its FedRAMP assessment process"

Need the 👍 from:
- [ ] @emaland 
- [ ] @NoahKunin 
- [ ] The authors, @mogul, @dlapiduz @brittag


brittag:
Wonderful, thanks! A few notes:

* In the title, "Cloud.gov" should be "cloud.gov" (we're slowly standardizing on cloud.gov as always lowercase).
* "documentation *with technical details" has a stray asterisk.
* I just noticed that "but here are more details about the two big categories our work fall into right now" should start with "and" instead of "but" (for a more positive tone). Also, this should say "falls into" instead of "fall into". (My mistakes, sorry.)
* "cloud-gov-inquires@gsa.gov" should have "inquiries" instead of "inquires" (for both the link text and the email link). (Also my mistake, I think!)

brittag:
Also the ship image can be wider/bigger if that would look nicer on the blog - I did a quick crop from the Flickr image, and it'd be fine to recrop it.

gboone:
@brittag Do you have a link to the Flickr image? I just downloaded the image from the Google Doc.

brittag:
@gboone Here you go: https://www.flickr.com/photos/usnavy/8428935049/ (it was a link in an older version of the doc, but it was a little subtle). It'd be nice for the image to link back to that Flickr URL too, for credit (even though it's public domain). :)

PR #1814, opened 2016-07-24
Corrected link to Stakeholder Interviews Method Cards

oghaffari:
It was missing the "discover" folder in the URL.


emileighoutlaw:
Thanks for the catch! Merging now.

PR #1821, opened 2016-07-28
Publish Bret's profile

gboone:


mogul:
Looks like it lost something at the end of this sentence:

> I had to learn how government approaches technology in traditional ways in order to think about how cloud.gov.

awfrancisco:
Going through the revision history, it looks like that's how the sentence was when Mel first transcribed it. Do you have an idea of how you'd like to end that sentence? We could also remove that sentence.

PR #1841, opened 2016-08-11
Adding an additional resource to open source post

brittag:
I'd like to add another useful link to "Facts about publishing open source code in government", based on discussion in #wg-opensource. While adding this link, I noticed that two list items had been merged into one item during the formatting process, so I also split those two list items.

cc @wslack


awfrancisco:
Looks good.

PR #1849, opened 2016-08-25
Micro-purchase: Do one thing well

gboone:
Publishes the post about the core architectural philosophies of the micro-purchase project.

cc: @18F/blog, @cmc333333 @jessieay @harrisj @adelevie


cmc333333:
This looks good, so per @gboone I'm merging

PR #1870, opened 2016-09-01
Copyediting for 18F style

nicoleslaw:


gboone:
Oooh, good catch. Thanks!

Issue #1608, opened 2016-03-31
Newsletter signup email input should be type 'email'


patrickcate:
I just signed up for the 18f newsletter at the bottom of: https://18f.gsa.gov in iOS Safari. I noticed for the email field the type is set to `text`. Changing this to `type="email"` would greatly improve usability, particularly on mobile devices where it brings up the email specific keyboard.

gboone:
I think you fixed this in #1612, right? Thanks for the contribution and feel free to re-open if that doesn't full address it.

PR #1894, opened 2016-09-13
Puts the newsletter link back

gboone:
Addresses #1608 

If you're sent or click on 18f.gsa.gov/#newsletter you get taken to the bottom of the page. I don't know that this is our best ever solution for the newsletter sign up. Particularly, I think it'd be easy to misunderstand what you're supposed to do at the bottom of the page depending on the context that sent you there. We'll leave #1608 open until we can test this and/or design a better interaction.


awfrancisco:
Awesome. Thanks for the temporary fix. Happy to discuss/research the best long-term fix.

PR #1920, opened 2016-10-03
Publish: Using RFPs to make every dollar count

gboone:
@MCHopson @18F/blog


MCHopson:
Looks good to me let's let it fly!

PR #1922, opened 2016-10-05
fix typo: "We delivering" -> "We deliver..."

jseppi:


awfrancisco:
Thanks for the catch!

Issue #1896, opened 2016-09-14
Automate accessibility scanning


gboone:
We have successfully setup `pa11y-crawl` to check the site for accessibility issues. Integrating it into our Circle CI build is doable but increases the build time to about an hour. This is because Circle builds the site fresh every time it pulls in the content, so even if only one page changed in the pull request, `pa11y-crawl` will scan every page in the site. Every content PR is going to take at least an hour (and fractions of seconds longer with ever new post), even if that PR changes `teh` to `the` on one page. 

Accessibility scanning would be great to do on every change. If we could do it immediately, even better but as far as I can tell there's not a great way to get Circle to only run against only the changed stuff. Perhaps there's a reduced set of posts and pages we could try to target? Circle has the ability to do [nightly builds](https://circleci.com/docs/nightly-builds/), this may be an option.

gemfarmer:
@gboone: @afeld, @mugizico, and I had a [Slack conversation](https://18f.slack.com/archives/g-devops/p1474991526000189) where we discussed tackling this accessibility compliance via the [Concourse Compliance Toolkit](https://github.com/18F/concourse-compliance-testing/issues/109) instead of on a per repo basis. Some other projects are already doing this via [continua11y](https://continua11y.18f.gov/), but given the large number of static files in this repo, I think we will need a different solution. 

We would like to set up a nightly build that will report the repo's compliance, but want to make sure that we can do that without running into billing issues.  Do you think that it will be ok to do this work?

gemfarmer:
(1) adding `pa11y-crawl` to Compliance Toolkit 
(2) adding https://18f.gsa.gov (or https://federalist.18f.gov/preview/18F/18f.gsa.gov/master/) to compliance viewer

gboone:
@adelevie mentioned something about running `pa11y-lambda` on AWS lambda once. Maybe that's a thing we can use?

gemfarmer:
After conversation with @afeld:

- [ ] set up concourse compliance viewer locally
- [ ] set up concourse backend tool locally
- [ ] set up 18f-site specific job with `pa11y-crawl`
- [ ] accomodate _all_ projects
    - [ ] make job task target project-specific url
    - [ ] make task run nightly
    - [ ] add configuration for project to opt-out of task

gemfarmer:
A repo specific pa11y scan has been added as of #1924. An enhancement to that branch could be to only scan urls that changed in that PR!

gboone:
[...]

PR #1923, opened 2016-10-05
[WIP] Incorporate pa11y ci

gemfarmer:
WIP 

#1896 

@mugizico Instead of waiting for Aidan to get to the concourse scanner, I thought we might consider solving this on our end first. With `pa11y-ci` I don't think it should take terribly long.


gemfarmer:
Wrong branch

PR #1935, opened 2016-10-12
Update ruby version to 2.3.1

coreycaitlin:
I worked with @gemfarmer to update the ruby version to 2.3.1, because of some recurring issues with 2.2.3.


coreycaitlin:
This also includes some updates on `master`, because I branched off master instead of dev initially.

PR #1945, opened 2016-10-14
Docker support(Update)

mugizico:
[WIP] and update to https://github.com/18F/18f.gsa.gov/pull/1688

 * Based on this  [Gitlab forum post](https://gitlab.com/gitlab-org/gitlab-ce/issues/14983) added UTF-8 language support in `Dockerfile`  to fix `jekyll 3.3.0 | Error:  invalid byte sequence in US-ASCII`  but it doesn't seem to work. I am thinking that it has to do with bundler version discrepancies. 

cc @gboone @gemfarmer @toolness

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM ruby:2.3.1
+
+# set locales 
+RUN  apt-get update >/dev/null && \
+     apt-get install -y locales >/dev/null && \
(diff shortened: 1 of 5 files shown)


mugizico:
https://github.com/18F/18f.gsa.gov/pull/1945/commits/fb0904f3853a1b3ae0adc210d9f8eda3a6cc2030 fixes enconding error.

Issue #1933, opened 2016-10-12
Make issue and PR templates


gemfarmer:
Make templates for new issues and PRs. This should help standardize things internally and for contributors

From EITI:
- [ISSUE_TEMPLATE.md example ](https://github.com/18F/doi-extractives-data/blob/dev/ISSUE_TEMPLATE.md) – add criteria to define when issues are scoped correctly. Micropurchase has some great examples here.
- [PULL_REQUEST_TEMPLATE.md example](https://github.com/18F/doi-extractives-data/blob/dev/PULL_REQUEST_TEMPLATE.md) 

This issue is done when:
- [x] we have a new issue template file
- [x] we have a new PR template file
- [x] related documentation

PR #1959, opened 2016-10-18
Issue and pull request templates

gemfarmer:
For #1933 

This PR makes templates for new issues and pull request templates. These templates will populate the content field when an issue or pull request is made. 

The issue for this PR says to add 'associated documentation', but I'm not sure where that would go, as it is fairly self-evident.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,15 @@
+## Expected Behavior
+
+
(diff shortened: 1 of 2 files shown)


coreycaitlin:
I agree that this is pretty clear — I don't think it's terribly useful to add documentation in another spot.

gboone:
Sounds like we're ready for a merge, then!

PR #1994, opened 2016-10-28
Changed second to next in line 134

Dahianna:
Changed second to next in line 134
@18f/blog
@awfrancisco


gboone:
Looks awesome!!! Thanks @Dahianna

Issue #2009, opened 2016-11-01
Author links broken


coreycaitlin:
## Expected Behavior

From any blog post author list, clicking on an author's name should take the user to a URL like `https://18f.gsa.gov/author/christopher-goranson`

## Actual Behavior

Instead, that link is being redirected to a URL like `https://18f.gsa.gov/site/18F/18f.gsa.gov/author/christopher-goranson/`, which 404s.

## Steps to reproduce the behavior
- [ ] [Go here](https://18f.gsa.gov/2016/10/26/writing-for-web-easy-writing-for-users-not/)
- [ ] Click on an author's name

Note: these links work locally AND in Federalist previews for the dev branch, which is why we didn't catch this. I suspect this has to do with how we're handling the baseurl?

## This is issue is done when:
- [x] Author page links work

mugizico:
I am trying to reproduce the behavior but the `Go here` link and clicking the author's name  is actually taking me to `https://18f.gsa.gov/author/christopher-goranson`

coreycaitlin:
Yay! This seems to be fixed with #2010 — closing out the issue.

coreycaitlin:
Reopening because this has recurred somehow 😕

coreycaitlin:
As far as I can tell, this is fixed with #2041 🎉

PR #2010, opened 2016-11-01
Fix author bug

gemfarmer:
I'm making the PR to `master` _because_ the bug is only on master

Fixes issue(s) #2009 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/debug-author)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/debug-author/)

It is hard to be 100% certain that this change will fix the bug seen in #2009, but I'm confident that it will. I think that the issue that we were seeing here is that URLs in production are constructed by combining a site `url` with a site `baseurl`. It is assumed that `baseurl` is an empty string, as we set it in _config.yml, but Federalist may be updating it. 

In this set of changes I'm 
- screening for the renegade `baseurl` string `site/18F/18f.gsa.gov`
- setting absolute URL in production if that text snippet is detected, otherwise restoring basurl to an empty string `""`
- On federalist, this should still work, because it will still be using the baseurl as it is currently.

/cc @coreycaitlin


gemfarmer:
@coreycaitlin @jamiealbrecht Did this fix the bug for either of you?

PR #2039, opened 2016-11-10
Redirecting

gemfarmer:
[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/redirection)

I think the issue that we were seeing with redirecting issues in production are the result of the following two factors:
- the `redirect_from` front matter properties contained paths that didn't end in forward slashes. This leads to files like `aaron.html` instead of `aaron/index.html` in `_site`, which is needed for federalist to work properly
- The `redirect_from` gem is likely creating the absolute url that it is redirecting the page to by looking at the configuration `site.url`, so it is getting `18f.gsa.gov` on federalist, instead of `federalist.18f.gov`. This _shouldn't_ be an issue on the live site. IMHO it isn't necessary to debug a redirecting issue if we are only seeing it in our staging environment. Especially because it will probably take a lot of extra time/work

NOTE (tldr): this is fixed locally and in production 🎉  for `team` --> `author`, but not for Federalist preview branches 😢 

/cc @mugizico

### DIFF
--- a/_authors/18F.md
+++ b/_authors/18F.md
@@ -9,4 +9,5 @@ state:
 github: 18F
 twitter: 18F
 team: 18F
+redirect_from: /team/18F/
 ---
--- a/_authors/aaron.md
+++ b/_authors/aaron.md
@@ -10,5 +10,4 @@ github:
 twitter:
 team: 18F
 redirect_from: /team/aaron/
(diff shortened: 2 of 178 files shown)


gemfarmer:
LGTM

Issue #2023, opened 2016-11-07
Create subnav and simplify nav management


coreycaitlin:
We have several sections of content that will need sub-navigation when we migrate them into the site proper (including Join 18F and the Partnership Playbook).

This issue is for implementing the design from #1954

coreycaitlin:
Blocked for implementation until we have the mobile design.

gemfarmer:
@ericronne is it fair to say that the mobile design is good to go? It sounded like we were just going to bootstrap with the USWDS

coreycaitlin:
Mobile design is in #2030 🎉 

Let's start with the simplest option here (i.e. WDS).

coreycaitlin:
Outstanding question: should this be a series of pages, or one long-scroll page?

coreycaitlin:
I've done a cursory content inventory, and I think our instinct that this should be a series of separate pages was the right instinct. Let's treat this as a traditional group of pages, rather than a single long-scroll page.

coreycaitlin:
Changed the title to reflect refactoring that needs to happen here

coreycaitlin:
Closing this out, now that we have subnav capabilities! We'll create a separate issue for pulling in content once we have our first section ready.

PR #2043, opened 2016-11-14
Subnav and navigation management

gemfarmer:
Fixes issue(s) #2023 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/subnav)

[:sunglasses: PREVIEW /hire/](https://federalist.18f.gov/preview/18F/18f.gsa.gov/subnav/hire/)

- [x] add subnav template
- [x] add mobile template


_sass/_core/typography.scss
```diff
@@ -6,6 +6,21 @@ body {
   font-size: $base-font-size;
 }
 
+html, body, div, span, applet, object, iframe,
```
gemfarmer: Normally I would agree, but in this instance, consolidating them makes sense and is more readable.

_plugins/utility.rb
```diff
@@ -0,0 +1,65 @@
...
+      # end
+      match
+    end
+
+    def check_type(value)
+      type = value.class
+      puts '---------------------'
+      puts "#{value} is a #{type}"
+      puts '---------------------'
+      binding.pry
```
gemfarmer: removed

gemfarmer:
Ok, ready for review @coreycaitlin @gboone @mugizico!

There are a ton of hound errors, but I am actively dismissing all of the remaining ones. 

One error is telling me to put selectors on multiple lines, which I would normally agree with, but in this instance, I'm not about to put 50 selectors on different lines. That would be completely unreasonable.

The other errors concern selector depth. This is frustrating, because Ideally I would create selectors with limited depth/specificity, but given that we are overriding the USWDS, I need to match their selector depth to properly override their styles. 😢

PR #2074, opened 2016-11-28
Update Modular procurement for state local government post

gboone:
@zacharycohn is this everything? I think I got it all.


gboone:
Zac says it's good. Merging.

PR #2076, opened 2016-11-29
Add version to config (v1.1.1)

gemfarmer:
Adds a version to the config file so that we can keep track of what version the code is at. I started at `v1.1.0` thinking that we were probably at `v1.0.0` when the current team started in October. Does this seem like an appropriate version?

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/version)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/version/)

/cc @gboone

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,6 +11,9 @@ logo: /assets/img/logos/18F-Logo-Bright-S.png
 feature_image: /assets/img/feature-background.jpg
 tag_dir: tags
 
+# app version number
+version: v1.1.0
+
 analytics:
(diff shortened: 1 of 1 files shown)


coreycaitlin:
Do we want the versions to line up with the [release log](https://github.com/18F/18f.gsa.gov/releases) at all?

gboone:
Let's start tacking on these version markers when we cut releases. It'll make them easier to follow.

PR #2082, opened 2016-11-29
Adds the code.gov update post

gboone:
Publishes "Curate the best thinking available"

Notes:
* I opted to go with text for the big block of JSON rather than a screenshot. I can easily swap it in if it's more appropriate to use the image. I previewed it in Federalist and it looks OK to me, just slack me if it needs to change.
* I changed the date to preview the post so that's what all the extra commits are about. If we merge this anytime in the morning it should publish correctly.

/cc @theresaanna @18F/blog


theresaanna:
The JSON block is perfect, thank you! It didn't occur to me to embed it!

gboone:
Thanks for the review!

Issue #2109, opened 2016-12-08
Search results include pages with empty titles


gboone:
## Expected Behavior
A search results page [like this one](https://18f.gsa.gov/search/?q=testing) should only show results with page titles and navigable page paths.

## Actual Behavior
Search results contain some pages with no title, and no clickable link.

## Steps to reproduce the behavior
- [ ] Do a search, any will do [but here's an example](https://18f.gsa.gov/search/?q=testing) and look for results like `1. ` 

## This is issue is done when:
- [x] All results of a search are clickable links

gemfarmer:
Ok, so after a bit more research, this appears to be caused by a few things
- files that weren't excluded in the build (contributing.md, for example) get indexed
- built redirect pages get built and are indexed. Adding `skip_index: true` to the frontmatter should prevent this, but several of the `redirect_from` pages don't have frontmatter.

gemfarmer:
Appears that adding `skip_index: true` to dead pages works, the site just wasn't building properly.

Also, I am adding the [results UI script](https://github.com/18F/jekyll_pages_api_search#results-ui-and-search-engine-options) to the javacript include so that we can filter the results. This will have the added benefit of allowing us to redesign the blog search results UI.

coreycaitlin:
I believe this is solved in [release 1.1.3](https://github.com/18F/18f.gsa.gov/releases/tag/1.1.3)

PR #2122, opened 2016-12-13
Blog search filtering

gemfarmer:
Fixes issue(s) #2109 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/post-filtering)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/post-filtering/)

- adds a [script](https://github.com/18F/jekyll_pages_api_search#results-ui-and-search-engine-options) to filter what search results show. This will also allow us to modify the style and structure of the [search page](https://federalist.18f.gov/preview/18F/18f.gsa.gov/post-filtering/search/) (ideas, @austinhernandez ?)
- adds `skip_index: true` to any pages that we are treating as redirect pages. From now on, if we add a `redirect_to` in the front matter, it should be accompanied by a `skip_index: true`, which will make it so the search engine doesn't index it.
- Exclude some markdown files from the build.

/cc @gboone @coreycaitlin


gemfarmer:
@gboone it looks like it is upset that I excluded CONTRIBUTING.md from the build. On the `/open-source-policy/` page, do we want to have a file download for CONTRIBUTING.md or a link to the repo?

gboone:
I think we could do fine with a link to the repo. That makes more sense to me.

gemfarmer:
@gboone merge conflicts fixed!

Issue #2110, opened 2016-12-08
Some author pages are duplicates and include all posts


gboone:
## Expected Behavior
Each author should have only one blog archive that displays only posts where they are a listed author. Any author listed in [`_authors`](https://github.com/18F/18f.gsa.gov/tree/master/_authors) who is not listed as an author should have an empty archive page or a 404.

## Actual Behavior
Some authors have a duplicate page that contain all posts in the site.

## Steps to reproduce the behavior
- [ ] Here are two example archives. [1](https://18f.gsa.gov/author/khandelwal/) [2](https://18f.gsa.gov/author/jayhuie/)
- [ ] For those authors [this is the correct archive for 1](https://18f.gsa.gov/author/shashank/) and the other should 404 or display a message saying that author hasn't written any posts.

## This is issue is done when:
- [ ] We've found all duplicate authors
- [ ] Archives show the correct posts.

gemfarmer:
@gboone what is the first example? I don't see a link to it

coreycaitlin:
You can find a bunch of these on this post and a few similar ones: https://18f.gsa.gov/2016/03/21/we-asked-100-of-our-coworkers-why-did-you-join-18f/

gemfarmer:
So it looks like whenever there aren't any records for an 18f team member who has an author page, we have been showing all the records!!! BOOOOO! :ghost:

PR #2125, opened 2016-12-14
Handle author with no blog posts

gemfarmer:
Fixes issue(s) #2110 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/authors-weirdness)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/authors-weirdness/)

[PREVIEW jayhuie](https://federalist.18f.gov/preview/18F/18f.gsa.gov/authors-weirdness/author/jayhuie/)
[PREVIEW shashank](https://federalist.18f.gov/preview/18F/18f.gsa.gov/authors-weirdness/author/shashank/)
[PREVIEW khandelwal](https://federalist.18f.gov/preview/18F/18f.gsa.gov/authors-weirdness/author/khandelwal/)

- Updates template so that authors pages with no associated blog posts are empty. This definitely needs to be changed. Maybe some basic information about the team member? @coreycaitlin 
- redirects `author/khandelwal/` to `author/shashank/`
- this significantly improves build times, like @gboone predicted. 🎉 

/cc @coreycaitlin


gemfarmer:
@coreycaitlin @austinhernandez for now, how do you think we should handle author pages that don't have any associated blogs. They are boring, so I'm inclined to not show them, but doing so would require adding `output: false` to all of the author pages that don't have blogs AND undoing this when we add an author's first blog post. 

We could automate this, but should probably avoid it if we can. A better alternative might be to create a default layout that all authors share? Maybe we can table it for now, and address it when a redesign is ready? The empty author pages aren't linked to anyway

austinhernandez:
@gemfarmer Where/when/why they are linked? 

Two follow ups:
- Are you suggesting a header or something that all authors have that has some basic info so that an empty page isn't so weird?
- Usually a empty page would need a redirect to eliminate user confusion (like a 404 error). Maybe all author pages could have a way to get back if we can't have specific "no content" text that says "We couldn't find any blogs by [name]. Try going back to 18F's homepage or blog."

gboone:
We don't link to these pages from anywhere within the site. Some people may have their old profile page linked to from an outside site. Not generating them will have the benefit of (slightly) speeding the build time and will only break links to pages without much useful content on them. I'm fine with the team going whichever way makes more sense or is easier to implement.

gemfarmer:
> when/where/why are they linked?
@austinhernandez there is often a link to an author page when someone is quoted or mentioned in a blog. The way that we are handling this now, _every_ author (who is specified in the `_authors` directory) has a page, but not all of them have actually authored posts. This is a legacy problem caused by combining what used to be a list of teammates, with a list of authors.

> Are you suggesting a header or something that all authors have that has some basic info so that an empty page isn't so weird?

exactly

@gboone I think this is spot on. I'll do some work to make sure that empty authors pages don't show.

> Usually a empty page would need a redirect to eliminate user confusion (like a 404 error). Maybe all author pages could have a way to get back if we can't have specific "no content" text that says "We couldn't find any blogs by [name]. Try going back to 18F's homepage or blog."
I think this is also a solid idea. This could be as straightforward as making our 404s more straightforward

_plugins/author_data.rb
```diff
@@ -0,0 +1,41 @@
+require 'yaml'
+
+module SiteData
+  class AuthorData
+    def initialize
+      @path = File.join(Dir.pwd, "_authors")
+    end
+
+    def update(author_file, key, value)
```
gemfarmer: this is fixed with 8cf5fcd

gemfarmer:
@gboone ready for review again!

gboone:
Looks good. The wrong `author` string for Shashank's redirected to 18f.gsa.gov/preview/<branch> which is strange but I think won't be a problem in production.

gemfarmer:
@gboone oh sorry for not making a note of that! I not entirely clear why Federalist preview branches direct to 18f.gsa.gov, but like you said, not an issue in production, so I agree that we are safe in production

Issue #2140, opened 2016-12-19
Tweak blockquote styling


coreycaitlin:

gemfarmer:
In.vision with block quote styling https://projects.invisionapp.com/share/3X9J87M4V#/screens/210062112_Project_Detail_V1

gemfarmer:
Block quote variations:

- **Basic blockquote**: Text that we want indented within a larger text block
- **Blockquote + Image**: 
- **Pull quote (see Austin's in.vision)**
- Lists of quotes: Ensure that blockquote styles look good en masse  (why did you join)
- List of quotes and images. (path to 18F)
- Image caption??: Maybe handle differently depending on length of the text

elainekamlley:
Examples of current blog posts that use block quotes: 
- [image caption](https://18f.gsa.gov/2016/08/02/jacob-harris-from-big-data-journalism-to-micro-purchase-platforms/)
- [blockquote + image](https://18f.gsa.gov/2016/08/26/mark-trammell-unlocking-the-true-potential-of-public-service/)
- [list of quotes](https://18f.gsa.gov/2016/11/25/what-18f-is-thankful-for-this-thanksgiving/)
- [list of quotes + image](https://18f.gsa.gov/2016/03/22/what-was-your-path-to-18f/)

PR #2146, opened 2016-12-21
Blockquotes

gemfarmer:
Fixes issue(s) #2140 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/blockquotes)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/)
[List of blockquotes](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/2016/11/25/what-18f-is-thankful-for-this-thanksgiving/)
[List of blockquote + image](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/2016/03/21/we-asked-100-of-our-coworkers-why-did-you-join-18f/)

- adds styling for `<blockquote>`and `<quote>` element in a number of layout situations, including:
-- single quote + image to be inline and indented
-- single quote to have left border and indented
-- list of quotes to adopt single quote styling 
-- 18F logo for quotes without images in list quote blog posts 

/cc @elainekamlley


_plugins/redcarpet.rb
```diff
@@ -28,6 +28,18 @@ def strip_dashes(str)
     end
   end
 
+  def block_quote(quote)
+    quotes = quote.split("\n\n")
+    if quotes.count > 1
+      blockquote = quotes.map do |q|
```
gemfarmer: I think it might be technically unnecessary, but definitely more clear

gemfarmer:
This is open for review!

awfrancisco:
The content doesn't seem to be properly displaying in [this preview](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/2016/11/25/what-18f-is-thankful-for-this-thanksgiving/). I know this post was listed in the original issue, but I don't think the original used block quotes. The quotes were just in normal body text. 

And I'm not sure if it's related to the blockquote styling, but a bunch of links seem to be broken in the [other post](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/2016/03/21/we-asked-100-of-our-coworkers-why-did-you-join-18f/), but the design looks much better. 

The new style does look good on the bottom of [this post](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/2016/08/31/kanban-for-government/)

coreycaitlin:
It looks like some issues with the `redcarpet` plugin are causing some blockquotes to spit out `blockquote.to_s` instead, as Andre noted.

gemfarmer:
@coreycaitlin @awfrancisco thanks for the reviews! I'm on it!

gemfarmer:
@awfrancisco I think both the posts you mentioned have been fixed:

- [blockquote.to_s issue](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/2016/11/25/what-18f-is-thankful-for-this-thanksgiving/)
- [markdown formatting issue](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blockquotes/2016/03/21/we-asked-100-of-our-coworkers-why-did-you-join-18f/)

awfrancisco:
Looks like there's still one cloud.gov link rendering as `[cloud.gov](http://cloud.gov/)` in the second post, but otherwise they look great.

gemfarmer:
@awfrancisco good catch! I looked through and fixed all other instances (on the three "we asked 100 of our coworkers..." posts) where we threw markdown into html blocks. Let me know if you find another instance like that!

awfrancisco:
Thanks, and will do!

coreycaitlin:
@gemfarmer redcarpet seems to be breaking up multiline/multi-graf blockquotes into separate blockquotes, which leads to this kind of thing:

screen shot 2016-12-30 at 11 02 03 am

Is it possible to treat that whole thing as one blockquote so it looks like this instead?

screen shot 2016-12-30 at 11 10 16 am

gemfarmer:
@coreycaitlin hmm 🤔. So this looks like an unintended consequence of trying to solve a different problem. Previously, redcarpet was taking separate blockquotes and merging them. So something like this
```
> blockquote 1

> blockquote 2
```
would render as

> blockquote 1
blockquote 2

instead of 

> blockquote 1

> blockquote 2

The way that we forced redcarpet to do this probably wasn't thorough enough. Back to the drawing board. 

cc @elainekamlley

gemfarmer:
@coreycaitlin ok, I think this last fix should wrap things up nicely. Instead of tinkering with the `redcarpet` renderer and extensions, I tried to upgrade to `kramdown` and I think the migration went _really_ smoothly. _Wayyy_ smoother than anticipated, really 💃 

Check it out here: 	9e330cd

kramdown is the default supported renderer for Jekyll 3.0.0 and above, so we are in good company there. Some benefits of kramdown:
- no blockquote weirdness 👏 
- we no longer need the extension `with_toc_data` to force headings to render with associated dash delimited id slugs. 👏 
- we no longer need the `smart` SmartyPants smart quotes extension 👏 
- it renders code blocks by default, and since we have a `rouge` stylesheet, there should be no migration issues here 👏 

I'm going to be clicking around to different pages thoughout the site, but so far, it looks like this is our solution!!! 🍾 🎆

PR #2193, opened 2017-01-11
The best way to build big is to start small

gboone:
Publishes this post by Kane!


gboone:
These accessibility issues have been resolved. The problem was I had the `src` URL wrong for the images.

Issue #2195, opened 2017-01-12
Better support for image layouts


gboone:
As a blog editor, I want an easy way to position and resize images in a blog post, so that I can make complex diagrams and wrap text around small images.

The images on a post like the one we posted today (will add a link when live) need to be handcrafted in order to float inline with the text. The result is unnecessary inline CSS, mixed markdown and HTML in blog posts, and repetitive code that is easy to get wrong. Because some errors require the author's feedback (which requires waiting for the Federalist preview to publish), these specialized layouts can delay publication by hours.

Related to #2185.

Issue #2185, opened 2017-01-09
Better shorthand for adding captions to blog/page images


gboone:
As a blog author, I want an easy shorthand for creating images with captions, so that I can spend less time converting google docs to blog posts.

When we have an image that doesn't need a caption, we can use the normal markdown shorthand `[alt text](path/to/image)` but if we do need a caption, we have to either craft a `<figure>` tag, or offset a seemingly unrelated paragraph in italics. It'd be great if we had a shorthand that could fill in the properties of `<figure>` and add a `<figcaption>` for us.

gemfarmer:
This is closed as of #2265 !

PR #2225, opened 2017-01-21
Blog template and image classes

coreycaitlin:
Possible approach to #2185 and #2195:

- Add a simple template in the `_posts` folder to serve as a reference for common Markdown needs and `figure` syntax
- Add 3 classes for blog image positioning: `blog-figure-left`, `blog-figure-right`, and `blog-figure-center`

This isn't terribly fancy, but it does provide a reference point to help take the guesswork out of putting images in blog posts, and give us the tools to avoid inline CSS when positioning images in posts.

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/blog-template)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/blog-template/)

@gemfarmer — Does this seem like a reasonable approach? Should we do something more robust? Is it weird to have the template file in the `_posts` folder, or should we have a separate place to keep template/reference files?

/cc @gboone @awfrancisco — does this help at all?

### DIFF
--- a/_config-blog.yml
+++ b/_config-blog.yml
@@ -360,7 +360,6 @@ exclude:
 - _posts/2016-12-02-presidential-innovation-fellows-now-accepting-applications-spring-2017.md
 - _posts/2016-12-12-connecting-culture-change-at-18F-and-PBS.md
 - _posts/2016-12-13-lowering-risk-of-experimentation-open-source-code.md
-- _posts/0-reference-post.md
 - _posts/2016-12-14-how-to-run-an-efficient-meeting.md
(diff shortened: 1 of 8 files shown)


gemfarmer:
@coreycaitlin this looks good. The only css feedback that I would give is to create less specific classes.

We can add this to the `_core/utilities.scss` file:
```scss
.align-center {
  @include clearfix;
}
```

Also, it looks like USWDS also has two classes, `.align-left` and `.align-right`, that do what you are doing with `.blog-figure-left` and `.blog-figure-right`.

Issue #2221, opened 2017-01-19
Remove title and description from right arrow image on "read more" link


maya:
The right arrow image on the "read more" link is decorative so I don't believe you need title and description. Announcing "Read more about [title of post]" will be the most helpful text for someone using a screenreader.

"about [title of post]" can be added as screenreader only text, which I believe is what it was like before, but @gboone may remember better.

/cc @nickbristow

gemfarmer:
@maya Because these arrows are used site wide to indicated 'read more', do you think a compromise could be to make the title `Read more` instead of `right arrow image icon`? Or do you think it is still confusing having that help text in conjunction with the other content in that card anchor tag?

maya:
@gemfarmer nope, bc that would be duplicate content bc the link already says "Read more." https://www.w3.org/WAI/tutorials/images/decorative/

nickbristow:
@gemfarmer yeah, I agree with @maya "arrow right" doesn't add anything and if there is text, its duplicate content and not needed.

gemfarmer:
@maya @nickbristow that tutorial is helpful. I went back through using VoiceOver and I see what both of you mean about the duplication. In that audit, I found the Featured Projects card links even more troubling. Do either of you know an effective strategy for putting a link around an element that has a lot of content without reading all of the internal content?

maya:
@gemfarmer that's a good question! I asked in the [a11y slack](https://web-a11y.herokuapp.com/) and got this answer:

>Personally I just add a click event handler on the whole card but don't make it a tab stop and then just focus on the real link for keyboard/screen reader users. just treat the click event handler as a redundant way of doing the same thing as the link

nickbristow:
@gemfarmer yeah, having a ton of content in a link isn't really a great for accessibility. Click handler on the card, and just having the link be focusable is probably the way i'd go

PR #2231, opened 2017-01-23
Accessibility updates to arrow SVGs

gemfarmer:
Fixes issue(s) #2221 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/carat-trouble)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/carat-trouble/)

- Updates to our `arrow-right` and `arrow-left` icons. They now have `role="img"` which makes them presentational and not apparent while tabbing on a screenreader. Checked manually with ChromeVoxLite
- Added `sr-only` help text near arrows in place of arrow image descriptions.
- restructured the cards on the homepage. The outer element is no longer an `<a>`, but has a click handler that follows the link provided. I then added anchors to the title and `Read more` text
- Pulls in a few unrelated changes to what is excluded by the `./serve` build script. We don't need to include markdown and helper files in our development build!

/cc @maya @nickbristow @coreycaitlin


_includes/svg/icons/arrow-left.svg
```diff
@@ -1,40 +1,7 @@
...
-</g>
-<g>
-</g>
-<g>
-</g>
-<g>
-</g>
-<g>
-</g>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" x="0px" y="0px" width="512px" height="512px" viewBox="0 0 444.531 444.531" style="enable-background:new 0 0 444.531 444.531;" xml:space="preserve" role="img">
```
nickbristow: For this svg add the attribute `aria-hidden="true"` so the screen reader won't read the "left arrow" part

```
<svg ... aria-hidden="true"> 
```

_sass/_core/typography.scss
```diff
@@ -35,6 +35,7 @@ h4 {
 
 a {
   color: $color-medium;
+  cursor: pointer;
```
nickbristow: The whole card is clickable, I noticed a couple of areas where there wasn't a pointer. The top image and some white space between text

gemfarmer: 👍

_includes/svg/icons/arrow-left.svg
gemfarmer: I tried it out using the ChromeVoxLite just adding `role="img"` and it seems to avoid the `<title>` and `<desc>` text. Are there other screen readers that _need_ `aria-hidden`

nickbristow: Yeah when i used Voiceover it was still reading the arrow.

gemfarmer:
@nickbristow I made both of your suggested changes. Let me know if you think this is 👌 to merge!

_includes/svg/icons/arrow-left.svg
```diff
@@ -1,40 +1,7 @@
...
-</g>
-<g>
-</g>
-<g>
-</g>
-<g>
-</g>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" x="0px" y="0px" width="512px" height="512px" viewBox="0 0 444.531 444.531" style="enable-background:new 0 0 444.531 444.531;" xml:space="preserve" role="img" aria-hidden="true">
+  <title>Arrow left</title>
+  <desc>Icon image of an arrow pointing to the left</desc>
```
maya: Is there a reason to include `title` and `desc` tags in the SVG? /cc @nickbristow

nickbristow: yeah, screen readers can read that. Although putting aria-hidden='true' on them will hide them anyway. 

Its kind of a wash. At the very least its there if someone forces a screen reader to read them anyway.

_includes/card.html
```diff
@@ -8,21 +8,22 @@
...
+         style="background-image: url({{ site.baseurl }}{{ include.image_src }});"></div>
     <div class="card-banner">
       <p class="card-tag">{{ include.agency }}</p>
-      <h3 class="h4">{{ include.tagline }}</h3>
+      <a href="{{ card_link }}" tabindex="-1" class="card-link-tagline"><h3 class="h4">{{ include.tagline }}</h3></a>
       <p class="card-description">{{ include.description }}</p>
-      <span class="link-arrow-right">
+      <a href="{{ card_link }}" class="link-arrow-right" tabindex="0">
         Read more
+        <span class="usa-sr-only">about {{ include.tagline }}</span>
```
maya: I tested this in VoiceOver and it's reading it backwards ("about..." is first, then "Read more"):
screen shot 2017-01-23 at 3 19 20 pm

Does "Read more" have to be in a `span` to be read first, since the other text is in a span? /cc @nickbristow

gemfarmer: Hmm that's interesting. I'll try doing that and check with VoiceOver

gemfarmer: @maya mine reads `more about ...` either way 😕 

screen shot 2017-01-23 at 5 54 06 pm

maya: Oh that's interesting! I have no idea why it's doing that then on my machine. 🤷‍♀️

_includes/card.html
```diff
@@ -8,21 +8,22 @@
...
-{% assign project_type = include.project_type | default: 'none' %}
-
-<article class="card {{ wds_grid_class }}" data-bucket="{{ project_type }}">
-  <a class="card-link" href="{{ card_link }}">
-    <div role="img" title="{{ include.image_alt }}" class="card-image-bg" style="background-image: url({{ site.baseurl }}{{ include.image_src }});">
-    </div>
+<article class="card {{ wds_grid_class }}">
+  <div class="card-link" onclick='window.location = "{{ include.link }}";' tabindex="-1">
+    <div role="img"
+         title="{{ include.image_alt }}"
```
maya: So it seems like you set your alt image tags manually in the templates, but I just want to mention that you shouldn't be using "Image of..." bc that's redundant as it being an image is already announced to users with the `role="img"` property. More info here: http://webaim.org/techniques/alttext/

coreycaitlin: I believe I've addressed this in `9c36a35`, but we should make sure this is noted elsewhere as an accessibility training resource, since I know people are often writing alt-text on the fly (not just for this site, but across 18F).

_includes/card.html
gemfarmer: It could just be on my machine. If you feel motivated to check it with/without the `<span>` we could determine if it is how one of our machines renders it

_includes/card.html
coreycaitlin: j/k, this is covered in the accessibility guide!

gemfarmer: 👍 👍 to both of these. @coreycaitlin maybe in our example post we could make sure to phrase the alt text so that we don't get duplication like Maya is talking about??

Issue #2186, opened 2017-01-09
Content: About 18F page


coreycaitlin:
We need a page that explains the basics of 18F, including:

- our relationship to TTS and GSA (i.e. #2155)
- our history
- funding
- more

coreycaitlin:
Source content from the press page:

> 18F is a growing group of technologists, designers, and researchers from around the country who are committed to making government services simpler and easier to use. We put the needs of the American people first, we work in the open to make our products stronger, and we are design-centric, agile, open, and data-driven.

> 18F projects adapted by other organizations

> We publish our code repositories on GitHub and encourage the public to adapt our code and guides for their own projects.
> 
> Here are some examples of how our projects have been adapted by others:
> 
> - A number of governments have used/adapted the code that powers analytics.usa.gov
> - The District of Columbia adapted 18F's Guides Template into their own, which they are using for things like the DC Government Open Source Guidelines.
> - Lovely, a real estate startup, created their documentation platform by adapting the code created for 18F's Hub. The Hub helps 18F organize our information and explore the connections between team members, projects, and skill sets. It serves as the go-to place for all team information.
> - Pulse was forked to check German government domains for HTTPS.
> - Many sites have adopted the Draft U.S. Web Design Standards.
> - AWS Broker for Cloud Foundry is in use by Australia's cloud.gov.au.

> 18F's relationship with the U.S. Digital Service
> 
> 18F and the U.S. Digital Service are distinct groups working toward the same vision of better digital government. 18F is a digital consultancy in an independent agency (GSA) and provides products and services to federal entities on a fee-for-service basis. The U.S. Digital Service is part of the Executive Office of the President and the United States CIO, supporting and coordinating the work of various digital service teams housed in agencies across the federal government.

> Basic information on 18F’s history
> 
> - GSA launched 18F on March 19, 2014. The name is an abbreviation for GSA’s address — 1800 F St.
> - The 18F team started with 15 full time staff. Staffing has grown to over 180 as of May 2016, as 18F’s product and client work has increased.
> - Our goal was (and is) to transform the way the U.S. government approaches problems. Read more about our evolution in the blog posts from our first and second anniversaries.

> How 18F Works
> 
> 18F works in partnership with agencies across government on a reimbursable basis. We deliver public-facing services via web applications, data and service Application Programming Interfaces (APIs), and platforms. 18F operates using three basic models:
> 
> - For you: Building solutions for an agency
[...]

PR #2262, opened 2017-02-06
Content revisions

coreycaitlin:
Fixes issue(s) #2186 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/content-revisions)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/content-revisions/)

- Adds some content to the press section
- Fix a broken link on the join page
- Add image list to press section

Still needs:

- check to see if we have the right cloud.gov logo
- refine image titles

/cc @awfrancisco


coreycaitlin:
hey @gemfarmer any idea why we keep getting those errors from `accesslint-bot`?

gemfarmer:
@coreycaitlin yes, I do. To make it possible for accesslint-bot to lint at all, we are only building the last three blog posts and their associated tags. If you are linking to old blog posts or tags not specified in any of the last three blog posts, the bot will think you are linking to pages that do not exist, because in that build, they don't.

I'm not sure what the best way around this is. I think it is useful to flag broken links, because that is helpful, but it is important for people to know that we will probably get some false negatives with this particular error.

PR #2263, opened 2017-02-06
Fixes headings and images for recent cloud.gov post

gboone:
cc: @jenniferthibault @18F/blog 

Replaces two of the images in the post and corresponding alt text. Also corrects the headings from being `**` emphasis to being `##` header mark up.


jenniferthibault:
I checked a preview sent over by @gboone , and looks good from my end. Thanks!

PR #2264, opened 2017-02-06
Mobile stickies

gemfarmer:
Fixes issue(s) #2264  

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/sticky-nav)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/18f.gsa.gov/sticky-nav/)

- changes `./serve`, `./serve-fast`, `./serve-blog` scripts to run at `/site/` instead of `/test/`. This seemed like a more appropriate nam
- Added mobile nav for sticky nav area
- Added mobile view for info block area

/cc @elainekamlley @austinhernandez @coreycaitlin


gemfarmer:
Blocked until we get a new chevron from @austinhernandez

austinhernandez:
@gemfarmer looks great! it looks like it's still being worked on because I'm seeing some drop down wonkiness

screen shot 2017-02-06 at 5 20 34 pm

I would say looking at this again, I don't think the "details" should open like a dropdown. That information can stay static on top and not be sticky on mobile. The stickiness works better for a navigation pattern.

gemfarmer:
@austinhernandez I think this is good to go!

coreycaitlin:
This is looking lovely! On the project pages (and just the project pages), can we make the right-hand sidebar not-sticky and not-collapsible on mobile? As @austinhernandez mentioned, it doesn't need to function like a nav.

gemfarmer:
Is there is there a reason why we wouldn't want it to be sticky? It seems like it would be an improvement even if we don't want to give it the visual affordance of a nav. But I could very well be off base there.  

I can totally remove the sticky-ness to it if we think that it would be too confusing, I just want to be sure that is what we want to do before putting in that time.

coreycaitlin:
Hm, I guess I was just thinking because it's not a nav, stickiness is a little confusing as an interaction pattern. It's just page content, so it makes sense to treat it like any other page content. I don't feel terribly strongly about it, but that's my hunch.

gemfarmer:
My hunch is probably sunk cost biased, so I'll side with you on this one.

PR #2298, opened 2017-02-15
remove broken dashboard links

coreycaitlin:
Removes an old link to the dashboard, which no longer works. It doesn't seem to affect meaning/sense at all.

/cc @gboone or @awfrancisco


gboone:
Looks like this is timing out, not failing, is that accurate? If so, I'm merging.

PR #2319, opened 2017-02-21
Update: Security and third party information post

gboone:
cc: @awfrancisco @vzvenyach


awfrancisco:
We're going with a different method to include the new info in the post.

Issue #1908, opened 2016-09-26
Bring join.18f.gov content into site


awfrancisco:
I'd like to replace https://18f.gsa.gov/join/ content with the content on join.18f.gov. Currently, 18f.gsa.gov/join basically just points to join.18f.gov, but when users get there they are presented with an entirely different design and no clear way back to 18f.gsa.gov. 

A review/update of the content on join.18f.gov would probably also be helpful. 

I believe this was mocked up in a way that preserved the join.18f.gov information architecture and used the Design Standards for the sidebar navigation. Will have to go look for that inVision file.

coreycaitlin:
Next steps:

- [ ] Find prior mockups — @awfrancisco, any luck?
- [ ] Content inventory to identify overlap and inconsistencies — @coreycaitlin 
- [ ] Meet to talk about how the content structure and design before we migrate it — @ericronne and @coreycaitlin

awfrancisco:
Added a step about talking to Talent. Happy to facilitate this.

coreycaitlin:
Created issue #1954 for the design chunk of this

coreycaitlin:
@rooneywp pointed out that it would be helpful to write a bit more about what we're trying to accomplish on the Join page. Here's my initial take:

- This page should give people the information they need to successfully send in their application to 18F (in conjunction with the PD or job posting, wherever that is)
- This page should also give people enough information about the remainder of the process that they have some reference point and understand what is normal.
- This page should be a welcoming, accessible introduction to some of the complexities of getting a job in the federal government.

We *don't* expect this page to accomplish these things:

- Telling people exactly what will happen or what order; this can change in each hiring round, and can be unpredictable from role to role
- Giving guidance about specific roles or position descriptions
- Duplicating content that can be found elsewhere

This page should be — at least — comparable to the information you can find about applying to private companies, and we fully understand and expect that applicants will have further questions to discuss with their recruiter or contact at 18F.

We should avoid content that changes frequently or is likely to become outdated, because we have limited resources for maintaining this content.

Anything else, @rooneywp @fureigh @awfrancisco ? Anything else we should look at from the old repo?

awfrancisco:
Those all sound good. Design wise, I think it should be immediately clear whether we have open positions (and what those are) and that any questions should be directed to join18f@gsa.gov.

awfrancisco:
[...]

Issue #2290, opened 2017-02-10
Add "go to top link" to sidenav


coreycaitlin:
We should add a link to the header on the sidenav so people can jump back to the top of whatever page they're on. This'll take a little thinking re: how it should work in mobile.

PR #2326, opened 2017-02-25
Join content revisions

coreycaitlin:
See issue(s) #1908 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/join-revision)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/18f.gsa.gov/join-revision/)

/cc @kategarklavs @rooneywp


gemfarmer:
Architecture work tracked here: #2335

gemfarmer:
@coreycaitlin I think this is ready for you to give another work. I'm not completely content with how I left the lead, but it gives us more to gnaw on tomorrow 😺

gemfarmer:
@awfrancisco Totally. I think that is even an issue: #2290

awfrancisco:
How about that. We'll I'm good to go on this PR then.

gemfarmer:
@awfrancisco I pushed the changes. Lemme know what you think!

cc @coreycaitlin

awfrancisco:
When you click on Join 18F, the highlight moves to/remains on "how to apply". Is it possible to remove the highlight when you click join 18F so it's in the same state as when you initially load the page?

gemfarmer:
For sure. Good catch. I didn't notice that.

gemfarmer:
I can't remove the url hash without refreshing the page, which people generally don't like, but I can jump it to the top of the page without highlighting the top nav item!

awfrancisco:
That's good with me

gemfarmer:
The primary outstanding issue is what to put in the subnav header text? The position title seems a bit redundant, but I'm not sure what is better. `Position info`? 

screen shot 2017-03-01 at 2 57 30 pm

PR #2338, opened 2017-03-02
Upgrade jekyll_oembed to v0.0.3

gemfarmer:
[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/upgrade-jekyll_oembed)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/18f.gsa.gov/upgrade-jekyll_oembed/)

- upgrades `jekyll_oembed` to `0.0.3`
- This is important because previously, private (or otherwise blocked) URLs would fail a build. `0.0.3` fails silently, which is good, in this case.

/cc @gboone

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ group :jekyll_plugins do
   gem 'jekyll_frontmatter_tests'
   gem 'jekyll_pages_api'
   gem 'jekyll_pages_api_search'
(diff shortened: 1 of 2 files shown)


gemfarmer:
This is safe, and a needed improvement.

PR #2336, opened 2017-03-01
Remove images

gemfarmer:
Fixes issue(s) #2331 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/remove-images)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/18f.gsa.gov/remove-images/)

Changes proposed in this pull request:
- removes some unused images from the site using an [image checker](https://gist.github.com/gemfarmer/f3f2e35663b96cd3fa8d90b49e6216a0) that I made to search the entire built site.

Some images I directly removed, others I found, but did not check in:

```
assets/blog/peacecorps/image01.png
.jpg
assets/blog/peacecorps/image01.png
t.png
assets/blog/checklistomania/checklists.jpg
assets/blog/cloud-gov/cg-logo.svg
assets/blog/complexity-and-security/security-documentation-2.jpg
assets/blog/design/bbc-design.jpg
assets/blog/govtechhack/IMG_3996.JPG
assets/blog/https-everywhere/18f-https-mobile.png
assets/blog/https-memo/whitehouse.png
assets/blog/new-jekyll-site/production-webhook.png
assets/blog/peacecorps/image01.png
assets/blog/speaker-series/cfpb-talk-slide.jpg
assets/blog/tockingtime/tock2.gif
assets/blog/vote-usa-gov/vote-og.jpg
assets/blog/were-hiring/map.svg
assets/img/logos/gsa-logo-light.png
assets/img/logos/pif-navy.png
assets/img/logos/pif.png
assets/img/logos/tts-logo.png
```
**What to do with the above**? ☝️ 

/cc @gboone @awfrancisco @elainekamlley @coreycaitlin

PR #2345, opened 2017-03-06
Image check

gemfarmer:
Follow up to #2336 

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/image-check)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/18f.gsa.gov/image-check/)

@gboone, remember when I said that the images that we would be removing in #2336 were definitely safe to remove? I was wrong 82 times 😅 😞 ! I was only scanning the built site for the complete path relative to the root, not the path relative to the css directory, so any background url images would formerly have been excluded in my check.

This version is much more vetted, with those exceptions in mind. Fortunately we didn't merge to master with the remove images PR.

/cc @gboone


gemfarmer:
@gboone For this PR, I added all the previously deleted images and essentially ran

`grep -R ../img/[image_path] _site` for each image against the entire built site. Previously, I had ran `grep -R assets/img/[image_path] _site`, so a handful of background images that start with `../img` weren't getting caught. Running the recursive grep gives yields the same result as doing a Find all in sublime, so I'm not sure of a better way to ensure that these aren't false negatives. Then again, I don't want to delete images unnecessarily again. Are there any other paths that would be worth checking, that you can think of?

So far:
* `assets/img/`
* `assets/blog/`
* `../img/`
* `../blog/`

PR #2390, opened 2017-04-10
Unlists the Product Manager postiion (it's closed)

gboone:
[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/unlisting)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/18f.gsa.gov/unlisting/)

[Preview README for this branch](https://github.com/18F/18f.gsa.gov/blob/unlisting/README.md)

- Unlists the product manager job description (because it's closed now)

/cc @hestherchang

### DIFF
--- a/_includes/open-positions.html
+++ b/_includes/open-positions.html
@@ -2,11 +2,12 @@
   {% if all_positions.size > 0 %}
     {% assign published_positions = all_positions | where: "published", true %}
     {% assign open_positions = published_positions | where: "listed", true %}
-
-<h3>Open positions</h3>
-  {% for position in open_positions %}{% unless "listed" == false %}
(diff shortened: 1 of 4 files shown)


gboone:
@gemfarmer can you confirm that /join/ looks correct in this PR?

gemfarmer:
@gboone the /join/ page is _slightly_ off, but it is not the fault of your PR. I think we missed some logic when we were redoing the page last week.

Line 6 and 7 can be swapped [here](https://github.com/18F/18f.gsa.gov/blob/unlisting/_includes/open-positions.html#L6-L7) so that `Open positions` doesn't show up when there aren't open positions.

gemfarmer:
@gboone or, more accurately, we need to update the if/else statement so that the if is checking that `open_positions.size > 0` instead of `all_positions.size > 0`

PR #2408, opened 2017-04-27
Adds workflow to convert docx to markdown

gboone:
A simple Automator workflow that copies markdown converted from a docx to the clipboard.


afeld:
Whoa! You should split this to its own repository - it's broadly useful.

gboone:
@afeld good idea, I'm also open to ideas on

1. Easy ways this could be installed on computers (could 18F/laptop do that?) 
1. Ways to include it with new Jekyll sites
1. Improvements that would dump the content into a specific file (I've given that a few tries but Automator isn't the most intuitive thing for that.)

gboone:
Closing this in favor of [18F/jekyll-macos-services](https://github.com/18f/jekyll-macos-services)

PR #2418, opened 2017-05-10
Acquisition Technical Lead position

gboone:
Adds a performance profile and three PDs for the acquisition technical lead.


gboone:
To be published Monday morning (6/15), taken down after 6/19. The listing says no applications will be considered after 6/19 so we can probably leave it in place until the following Monday. That's what we've done in the past, anyway.

PR #2423, opened 2017-05-15
add description tag for acq tech lead posting

stvnrlly:
This adds the YAML to create a description metatag for the tech lead job posting (I hope).


elainekamlley:
@stvnrlly where are you looking to have the description render? Under the `technical lead` button? If so, the layout doesn't currently add descriptions. We could be more descriptive on the button.

stvnrlly:
I was looking to create a metatag that'll get expanded on things like twitter, facebook, slack, etc. I don't think this should change anything on the page itself.

elainekamlley:
Ah, I see what you're saying. Merging!

PR #2435, opened 2017-05-22
Take down job listing

gboone:
Position is closed, so let's stop advertising it.

cc: @hestherchang @awfrancisco


hestherchang:
thank you everyone!!

PR #2446, opened 2017-06-06
Fractal and Federalist for the USWDS

gboone:
Publishes the post. cc @awfrancisco @18f/blog


wslack:
@gboone can we add a line about the recently added feature that pushes builds back to GitHub Pull Requests: https://github.com/18F/federalist/pull/906

Something like

> Federalist also recently added a feature that pushes build statuses back to GitHub, so any pull request that Federalist can't build into a preview is flagged with a link to the logs showing the build error.

PR #2479, opened 2017-07-25
Fix external link to Federalist

zackseuberling:
The external link to the Federalist was not working because it did not use the protocol.

- Add https protocol to the external Federalist link

/cc @gboone

### DIFF
--- a/_styleguide/subpages/for-developers.md
+++ b/_styleguide/subpages/for-developers.md
@@ -4,7 +4,7 @@ subpage: For developers
 permalink: /styleguide/for-developers/
 ---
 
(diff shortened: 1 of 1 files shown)


elainekamlley:
@zackseuberling thanks for the catch! Since forks can't go through `circleCI` I can make that update on our end.

zackseuberling:
@elainekamlley I didn't know what the protocol was for outside contributors, so I did what I thought would be the least harmful

elainekamlley:
Sorry about the confusion. Thanks so much for the correction! I added it to PR #2482

PR #2484, opened 2017-07-28
Unpublish code of conduct (for now)

gboone:
Take it down while we work out some issues with the existing code of conduct.


awfrancisco:
Do we need to take the link out of the footer too?

elainekamlley:
Yea we do. I'll make that push.

PR #2499, opened 2017-08-11
Add line breaks where intended to gov resume format

tvanantwerp:
This PR modifies the example government-style resume to insert line-breaks where seemingly intended according to the original markdown file. Single line breaks in markdown don't automatically translate to `<br />` tags. To create the line breaks in the HTML, I added two single spaces to the ends of the lines where a line break should occur. [Two single spaces or more at the end of a line in markdown](https://daringfireball.net/projects/markdown/syntax#p), when rendered to HTML, creates the intended `<br />`.

This fix can be a bit tricky to preserve though, if anyone who edits the file has settings in their editor to strip trailing spaces on save.

## Before

image

image

image

## After

image

image

image

### DIFF
--- a/_join/index.md
+++ b/_join/index.md
@@ -146,9 +146,9 @@ There are many guides to building a government-style resume. Here are several re
 
 This guide shows how to format a government-style resume and what information to include:
 
-> **Name**
-> **City and state of current residence**
-> **Email address**
+> **Name**  
+> **City and state of current residence**  
+> **Email address**  
 > **Phone number**
 >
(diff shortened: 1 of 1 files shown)


mirrorstage:
Came here to write a PR on this, too. Good eye, @tvanantwerp! Hopefully won't result in too many one-line info blocks in submitted résumés...

elainekamlley:
Thanks for this catch!

elainekamlley:
Closing since fix is in #2539. Thanks @tvanantwerp!

PR #2520, opened 2017-08-28
Minor copyedits / typo fixes

nicoleslaw:
cc @awfrancisco Missed it before you merged. I think the GS-14 one also has an error under Qualifications. It says GS-13 there?

### DIFF
--- a/_join/open-positions/content-designer-gs14.md
+++ b/_join/open-positions/content-designer-gs14.md
@@ -1,10 +1,10 @@
 ---
-title: Content Designer GS-14 
+title: Content Designer GS-14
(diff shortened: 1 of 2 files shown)


elainekamlley:
I think GS-13 is accurate. I think to qualify for GS-14 positions you have to meet GS-13 requirements. @hestherchang can you confirm?

nicoleslaw:
Okay, the reason I noticed it is that the bullets are the GS-14 experience, but it makes it seem like they are GS-13: https://18f.gsa.gov/join/consulting-software-engineer-gs14/#qualifications

PR #2543, opened 2017-09-13
Fix tiny typo on jobs page

tdooner:
Maybe I'll find more typos as I apply for a job... but for now, here's
an accidental double-quote.


awfrancisco:
Thanks for the catch! CircleCI can't run on a fork, so I'll close this and make the update on our side.

PR #2613, opened 2017-12-11
updated email to jointts@gsa.gov

elainekamlley:
Fixes issue(s) # [joining-tts](https://github.com/18F/joining-18f/issues/355)

[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/email-update)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/18f.gsa.gov/email-update/)

- Changed email to jointts@gsa.gov

/cc @awfrancisco @JennMoran

### DIFF
--- a/_join/index.md
+++ b/_join/index.md
@@ -25,7 +25,7 @@ We’re looking for candidates passionate about our mission, with top-notch soft
 
 We’re a remote-first team with offices in DC, New York, Chicago, and San Francisco, and we have teammates working all over the country. [Read more about how we make our distributed model work]({{ site.baseurl }}/2015/10/15/best-practices-for-distributed-teams/).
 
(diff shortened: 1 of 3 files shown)


JennMoran:
Hey - So I was actually wondering if we could not show the email and just have a link to the email where is says  Talent Team so people would click on the word Talent Team and be taken to our email, just so they don't get confused when it says joinTTS and not 18F, or do you not think that is a good idea

awfrancisco:
I think it's good to put the email on the site so people can add it to their address book, save it for later, or otherwise know it without immediately clicking the link. Also, they are actually also joining TTS when applying for an 18F job. Have you had any confusion from people in the past who applied to non-18F jobs but interacted with the join18f email address?

PR #2695, opened 2018-03-14
Launches updated How we work and What we deliver pages to production

elainekamlley:
[CircleCI](https://circleci.com/gh/18F/18f.gsa.gov/tree/dev)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/18f.gsa.gov/dev/)

[Preview README for this branch](https://github.com/18F/18f.gsa.gov/blob/dev/README.md)

- `/how-we-work` includes new content on services we provide and adds projects to the bottom of the page
- `/what-we-deliever`: Project cards are broken up between services and products
- Project pages can have multiple GitHub repos
- Merged `hiring` tag into `join us`
- New `blockquote` styling for testimonials

/cc @awfrancisco


awfrancisco:
@elainekamlley It looks like the project cards on How we work use the subtitle field from the front matter instead of the title field as the header for the cards. For instance, the FEC card should be "Easy-to-use campaign data." Can you switch those to match the cards on What we deliver?

Issue #2752, opened 2018-05-10
Move to CircleCI 2.0


awfrancisco:
Email from Circle:

We’re writing to let you know that your project(s) https://github.com/18F/18f.gsa.gov have been running on Ubuntu version 12.04 within CircleCI 1.0. Ubuntu ended support for 12.04 more than a year ago. Continued use of 12.04 is not recommended. We will be sunsetting CircleCI 1.0 and removing all support for Ubuntu 12.04 on CircleCI on August 31st, 2018.

What’s at risk with continued use of Ubuntu 12.04?

Continuing to use this version leaves your projects susceptible to security vulnerabilities since Ubuntu Security Notices for this version no longer include information or updated packages, including kernel updates.

Next steps: 

- Determine if your project(s) will need to migrate to CircleCI 2.0. You should move all projects that you plan to continue building on CircleCI after August 31st, 2018.
- For any inactive projects that you don’t intend to migrate, visit the project settings page and click on the “Stop Building” button to remove the webhook.
- Next, migrate your projects to CircleCI 2.0 by updating your circle.yml to a .circleci/config.yml (Tip: Use [this script](http://go.circleci.com/R00Gf71Ml20020f0ZhHo8k0) to generate the basic configuration file for CircleCI 2.0)
- Choose the executor you want to run your builds on. See how [here](http://go.circleci.com/B1009Z2Hk000f0oM07iGfl2). 

Our Success Engineers are available to support your migration efforts and make sure you can successfully build on CircleCI 2.0. If you’re interested in connecting with a Success Engineer, please let us know.

eddietejeda:
I am looking into this with @adborden

PR #2757, opened 2018-05-14
Update circleci to version 2.0 and disable accesslint, which caused issues. Re-enabling accesslint in different pull request

eddietejeda:
- updates the config for circleci. this is a work in progress to see if these configurations work in our environment. 
- See issue #2752 

/cc @adborden @awfrancisco


.circleci/config.yml
```diff
@@ -0,0 +1,21 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/ruby:2.3-node
+    enviroment:
```
adborden: Typo: should be `environment`

.circleci/config.yml
```diff
@@ -0,0 +1,21 @@
...
+      ACCESSLINT_MASTER_BRANCH: dev
+    steps:
+      - checkout
+      - run: bundle install
+      - run: npm install
+      - run: npm run test
+      - run: npm run htmlproofer
+      - run: bundle exec rspec
+      - run: CODECLIMATE_REPO_TOKEN=${CODECLIMATE_REPO_TOKEN} bundle exec codeclimate-test-reporter
+      - run: ./serve-accesslint && bundle exec accesslint-ci scan http://localhost:4000/site
```
adborden: Since the accesslint-ci tool has moved to a service, I think we should remove this step and create a new issue to enable the accesslint-ci service instead.

Issue #2772, opened 2018-05-25
Add loading icon to search results page


eddietejeda:
The search results page has a multi-second delay before actual results are shown. The long delay looks like an error. This could be resolved by adding a loading icon.

- [x] Review code that returns results
- [x] Add loading icon

PR #2775, opened 2018-05-30
Add loading bar to search

eddietejeda:
The old results page returned a blank page before loading results. The long delay looked like an error.

Since the search interaction is managed by https://github.com/18F/jekyll_pages_api_search (and does not appear very active), this is a simple work around to deal with the problem right away. Not ideal, but it gets the job done.

How it works: We simply show a loading icon when the page loads and we use jquery to hide the icon when results are loaded. Fortunately, jekyll_pages_api_search has a method we can override called `renderJekyllPagesApiSearchResults` when results are retrieved. So I simple hide the loading bar when that method is called.

References #2772 

[PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/18f.gsa.gov/add-loading-bar-to-search/search/?q=blog)


awfrancisco:
That loading animation looks good to me. And interestingly, when I search for `software` I don't get the blank lines error that I do on the live site.

awfrancisco:
@eddietejeda Are we all good on the hound alerts on this one?

eddietejeda:
I actually need to hold off on this. I just found a bug when there are no results the loading bar never disappears. Will keep you posted.

eddietejeda:
This pull request is now dependent on this patch to https://github.com/18F/jekyll_pages_api_search/pull/44

Will hold off on merging this feature until that request is merged.

PR #2787, opened 2018-06-13
Remove extra word

gregjd:
Just a very small change — I noticed the word "is" was in there twice by accident, so I figured I'd submit a PR to address it. 🙂

- Remove a likely typo/extra word

(I deleted most of the PR template fields here as they look like they're not so relevant for this PR)


wslack:
Thank you @gregjd!

PR #2860, opened 2018-08-15
center-align figure captions

afeld:
I personally think the captions look better center-aligned, but this is just a suggestion!

## Before

screen shot 2018-08-15 at 12 19 11 am

## After

screen shot 2018-08-15 at 12 19 21 am


eddietejeda:
Good suggestion, but we'll pass on it.

PR #3018, opened 2019-05-23
Fix dot-voting link in "Centralization Gone Right"

chuckhoupt:
Sometime before 2019, the methods.18f.gov site refactored without installing proper redirects (reported in https://github.com/18F/methods/issues/405). The broken link should be fixed in anticipation of the fix to methods.18f.gov.


Dahianna:
Thank you @chuckhoupt, we'll make the changes. Also, I checked on the old link and a redirect already exists.

PR #3080, opened 2019-11-03
(Fix) Footer Accordion

magnessjo:
### Fixes issue(s)

Add expand/collapse functionality.
Modify HTML for hidden attribute (to display none)
Overwrite position absolute for accordion content.

Note: This does not include any style updates or animations.  

### Reproduce Steps

  1. Navigate to https://18f.gsa.gov/ with a window width below 768.  Scroll to the footer.  Try to expand or collapse accordions.  

### Visual Reference 

Screen Shot 2019-11-03 at 11 02 23 AM

/cc @relevant-people

### DIFF
--- a/_includes/navigation/footer-mobile.html
+++ b/_includes/navigation/footer-mobile.html
@@ -21,7 +21,7 @@
       <button aria-controls="amendment-2">
         Policies and Standards
       </button>
-      <ul id="amendment-2" class="usa-accordion-content usa-sidenav-sub_list">
+      <ul id="amendment-2" class="usa-accordion-content usa-sidenav-sub_list" hidden>
         <li>
           <a href="{{ site.baseurl }}/guides">Guides</a>
         </li>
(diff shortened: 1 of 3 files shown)


lsgitter:
@apburnes  to review the fix.

magnessjo:
I noticed a lint test was failing since it expected jQuery.  I rewrote in jQuery and added the vanilla below (in case you would prefer it).  

```
// Vanilla Content Loaded 
document.addEventListener('DOMContentLoaded', () => {

  // Footer Accordion
  function footerAccordion() {

    const containers = Array.from(document.querySelectorAll('.footer-mobile'));
    
    // Check for multiple entries 
    for(var i = 0; i <= containers.length - 1; i++ ) {

      // Scope each instance 
      (function(containerRef) {

        const buttons = Array.from(containerRef.querySelectorAll('button'));

        function showHide(button) {
          const parent = button.parentNode;
          const id = button.getAttribute('aria-controls');
          const element = parent.querySelector(`#${id}`);

          if(element.hasAttribute('hidden')) {
            element.removeAttribute('hidden');
            button.setAttribute('aria-expanded', 'false');
          } else {
            element.setAttribute('hidden', null);
            button.setAttribute('aria-expanded', 'true');
          }
        }

        buttons.forEach( (button) => {
          button.addEventListener('click', () => showHide(button));
        })

      })(containers[i]);
    }

  }

  // Load
  footerAccordion();

});
```

Dahianna:
@apburnes to ask amirbey about build issue

Dahianna:
Thanks for the footer code. We are actually working on a new footer that meets the new government modernization guidelines, and it'll fix this issue.  Closing for #3180

PR #3104, opened 2020-01-10
Point to Engineering Guide vice Frontend

ryanhofdotgov:
We've consolidated the frontend guide into an overarching engineering guide at https://engineering.18f.gov, and so this just does a URL swap.

I've retained the HTML/frontend-centric image; it's not perfect, but it gets the job done.


Dahianna:
I'll take a look and merge

PR #3252, opened 2020-09-30
fix link

wslack:
I still reference this post and a link needs updating!


Dahianna:
Thanks @wslack ! I'll accept the changes.

Issue #3317, opened 2021-01-08
Add autoprefixer to jekyll-assets


apburnes:
## Expected Behavior

Building the SCSS will autoprefix the CSS styles when building with `jekyll-assets`.

## Actual Behavior
Will add the vendor specific prefixes to css styles that apply to better support multiple browsers. Should fix #3290

## This issue is done when:
- [x] Install autoprefix gem and configure based on [`jekyll-assets` docs](https://github.com/envygeeks/jekyll-assets#css-auto-prefixing)
- [x] Output css is being autoprefixed

Dahianna:
Fixed by #3302

PR #3336, opened 2021-02-08
Add auto-prefixer

### LINKED ISSUE #3290: Broken icon on official gov site banner
## Expected Behavior
The banner should display a down chevron 
image

## Actual Behavior
Displaying a horizontal bar
image

I am guessing it has to do with this `mask` property
image

igorkorenfeld:
Adds in the autoprefixer gem, addressing #3317 . The settings are copied over from USWDS. Should close #3290 as well.

:sunglasses: &nbsp; [Preview](https://federalist-d703a102-c627-4805-8934-78f5bf4c97da.app.cloud.gov/preview/igorkorenfeld/18f.gsa.gov/ik-autoprefix/)

/cc @apburnes


admin/config.yml
```diff
@@ -12,6 +12,13 @@ public_folder: /assets/blog
 publish_mode: editorial_workflow
 logo_url: https://18f.gsa.gov/assets/img/logos/18f-logo.svg
 
+autoprefixer:
```
apburnes, line 15: I think this config needs to be in the `_config.yml`. This config is for the Netlify CMS and not Jekyll.

apburnes (commented):
Thanks Igor! Looks pretty good. I think we just need to move the autoprefixer settings into the Jekyll config.

admin/config.yml
igorkorenfeld: Oh my mistake, thanks for catching that.

igorkorenfeld:
@apburnes pushed up the updates. Just wanted to double check my understanding, the new gem does need to be in the general `Gemfile` in addition to the `GemfileFederalist` right?

admin/config.yml
apburnes: Awesome!

apburnes:
That's right @igorkorenfeld. We need the `GemfileFederalist` because one of the test dependencies cannot be installed in the Federalist build environment.  I think there is a more elegant way of requiring dependencies based on build type within the `Gemfile` but I have not had time to figure it out.

apburnes (approved):

Dahianna:
Thanks @igorkorenfeld !

PR #3357, opened 2021-02-25
Update theme scss

lauraGgit:
updates the files in the `_theme` to the version of uswds pulled in from the theme.

The question remains if we only want to surface the values were changing (mainly in typography and in color) or if we want to pull in the full theme styles?


peterrowland (approved):
Looks good to me, and we agreed in meeting today that we want to have all theme vars available for future devs. 

Hound really doesn't like comment art.

lauraGgit:
@peterrowland yeah, i updated the linter rules so that going forward hound will chill on the errors. thanks for taking a look

Issue #3376, opened 2021-03-09
What we deliver page review


igorkorenfeld:
## Intro Hero
Please see #3374 

## Intro text

![image](https://user-images.githubusercontent.com/52677065/110501872-a1205f80-80c8-11eb-85ce-a86ae9a86649.png)

- Line missing after the intro text
- The text looks like it’s floating and somewhat awkwardly contained, so I suggest we bump up intro text to `lg` size, and allow it to take up 9 columns. This may need to change across all of the pages
- Should have 5 units of space after intro text and then 5 units of space after the `hr`  (above “Case Studies”)

## Cards
- Since the cards have a drop shadow they do not need a boarder 
Current branch:
![image](https://user-images.githubusercontent.com/52677065/110498946-c8296200-80c5-11eb-9781-7be0200a150a.png)

Proposed:
![image](https://user-images.githubusercontent.com/52677065/110499062-e4c59a00-80c5-11eb-9a54-bce6e5054e10.png)

## Quote
- Missing hr line after cards
- Should have 5 unites of space after cards before the line and 5 units after the line before the quote
- Could we increase the line-height of the quote to 1.35 (`lh(3)`) ? Note that the large quote mark should stay aligned with the top of the quote
- Quote attribution should start with an em dash (—), currently it's a hyphen (-)

## Agencies 
- Decrease Padding above Heading to 5 units
- Each agency list item should have 2 units of space below it
- Wondering if for the U.S. Air Force and for the U.S. army we ought to use their respective seals, in place of the logos to be more consistent with the other branches and agencies? 
[Army seal](https://en.wikipedia.org/wiki/United_States_Army#/media/File:Mark_of_the_United_States_Army.svg),  [Air Force seal ](https://en.wikipedia.org/wiki/United_States_Air_Force#/media/File:Mark_of_the_United_States_Air_Force.svg)

## Call to action
- Same as issue #3367 

## Summary Checklist
**Intro hero**
- [x] Changed based on #3374 

**Intro text**
- [x] Rule is added back in, is `gray-cool-10 `(`#dfe1e2`)
- [x] 5 units of space after intro text and then 5 units of space after the `hr`  (above “Case Studies”)

To be discussed: 
- [x] text is `lg` and takes up 9 columns instead of 8

**Cards**
- [ ] No border on the cards
- [ ] Rule after cards, with 5 units of space between the cards and the line
- [ ] Rule is is `gray-cool-10 `(`#dfe1e2`)

**Quote**
[...]

Issue #3375, opened 2021-03-08
Contact page Review


igorkorenfeld:
## Intro Header
See #3374 

## Main content 
Link color is `#00547a` should be `#046B99` ( `#00547a` is not a brand color)

## Sign up

- Sign up button is slightly larger than the search box, making it look misaligned 
![image](https://user-images.githubusercontent.com/52677065/110390190-9b2c6f00-8033-11eb-8ace-6d1ef850ed36.png)

- Left side of the sign up button should not have rounded corners so it appears connected to the input, like the search button on a search box

![image](https://user-images.githubusercontent.com/52677065/110390800-87353d00-8034-11eb-815d-2745d3f75923.png)

I am wondering if this section should be a call to action like on the Guides page. Here is a mockup, would love to hear your feedback @Dahianna, @lauraponce, @lauraGgit 

![111_18FSite_Signup](https://user-images.githubusercontent.com/52677065/110390546-260d6980-8034-11eb-877b-057006d7eecc.png)

If not, separator line should be `gray-cool-10` (`#dfe1e2`), and should only be as wide as the text content like the rules on current [What we deliver](https://18f.gsa.gov/what-we-deliver/) page

## Summary Checklist
**Intro Hero**
- [ ] Changed based on #3374

**Link text**
- [ ] Link color is `#046B99`

**Signup**
Direction A
- [ ] Styled as in mockup, matching Call to Actions on other pages
 
Direction B
- [ ] Rule is `gray-cool-10` (`#dfe1e2`)
- [ ] Rule is only as wide as the grid

lauraGgit:
@igorkorenfeld would it be ok to use the existing grid layout of the banner that is on the guides banner cta?

igorkorenfeld:
I find that using the `media-block` container with a `.usa-media-block__body` and `.usa-media-block__img` works better for creating a visual relationship between the text and the image. So ideally I would switch both to that. _But_ if that isn't possible to use here,  or if there is another reason not to use it, then I think using columns and the grid is a good alternative.

lauraGgit:
happy to port it over to media block - do you know where in the uswds docs I can learn about that?

igorkorenfeld:
Awesome! I'm not having luck finding documentation for it. I see it being used on the example page template for the [landing page](https://designsystem.digital.gov/page-templates/landing-page/)

Issue #3368, opened 2021-03-04
Footer Update Review


igorkorenfeld:
![image](https://user-images.githubusercontent.com/52677065/110024357-30f69000-7cfc-11eb-9935-d85816d82739.png)

- The headers are just slightly smaller than the link text. They should be `lg` to be more differentiated. 
- The headers have an underline and a click pointer on hover, but they are not a link so there shouldn't be any interaction effect
- Links are `#00547a` and should be `#046B99`. (`#00547a` is not a brand color) 

## Summary Checklist
- [x] Headers are size `lg`
- [x] Headers are static text with no interaction effects. 
- [ ] Links are `#046B99`

divyanshdahiya:
i have the solution to this problem and i can fix it easily .

PR #3382, opened 2021-03-12
Footer/ WWD/ Contact Page: design review tweaks

### LINKED ISSUE #3374: Intro Hero Review
Pulling this out of #3367, since it affects every page except the home page, and I wanted to make sure the discussion isn’t lost. I'll refer to this component as "Intro Hero" because it seems like "Section Header" might be ambiguous. 

image

My original suggestion

- On the heading the background is `#162e51` which is not a brand color. Should be `#1C304A`
- The text is more “chunky” than our current site, despite being the exact same weight. I think this might be due to some “font-normalization” that USWDS does. We could compensate for that and  make it more readable by adding some letterspacing. 
Existing site:
image
Branch:
image

 
The other issue was about the size.
“Looking at other pages on the current site, I realized the H1 ought to be smaller. as well, as it's being used as a kind of pre-header text. Size should be ~2.3 rem, so something like lg should work?”

On our current site the name of the page, the `H1` is smaller than the “tagline” text below it. I believe the original design is trying to emphasize the description of the page content over the title. I think this works because the current page is also indicated in the navigation and the page name does not necessarily need to be overly emphasized. Because is only one `H1` on the page and it appears before the other headings, the `H1` makes sense semantically and in the logical page structure, and so I don’t believe that it causes any accessibility issues. 

Having said that, if we prefer it to be a big header, my main issue with the typography on the branch is that the tagline is only slightly different in size from the header. I’d like to differentiate it, in order to make clear that it’s not a continuation of the same text, and to make the difference seem intentional. So in that case I would propose we shift the weight to regular and try it at size `md`

lauraGgit:
**SEE the project board for which items are in progress/ done of the design label tickets.**

Incorporate HWW and netlify changes from existing site.

Close #3374

 ## Contact Page #3375 
**Intro Hero**
- [X] Changed based on #3374

**Link text**
- [X] Link color is `#046B99`

**Signup**
Direction A
- [X] Styled as in mockup, matching Call to Actions on other pages

  
## Footer: #3368
- [x] Headers are size `lg`
- [x] Headers are static text with no interaction effects. 
- [X] Links are `#046B99`

## What We deliver: #3376
**Intro text**
- [x] Rule is added back in, is `gray-cool-10 `(`#dfe1e2`)
- [x] 5 units of space after intro text and then 5 units of space after the `hr`  (above “Case Studies”)

To be discussed: 
- [x] text is `lg` and takes up 9 columns instead of 8

**Cards**
- [x] No border on the cards
- [x] Rule after cards, with 5 units of space between the cards and the line
- [x] Rule is is `gray-cool-10 `(`#dfe1e2`)

**Quote**
- [x] Style to new homepage designs.

**Agencies**
- [x] 5 units of space between heading and top of grey container
- [x] Agency list item has 2 units of space below

**Call to Action**
- [x] Background color should be `#1C304A`  
- [x] Text is lg size
- [x] Button has background color `#00CFFF` and font-color `#1C304A`.  (can follow the [ACCENT COOL COLOR ](https://designsystem.digital.gov/components/button/)example for states)


_includes/testimonial.html
```diff
@@ -0,0 +1,13 @@
+<section class="usa-section grid-container testimonial">
+    <div class="grid-row grid-gap">
+      <div class="tablet:grid-col-8 testimonial__body">
+        "{{ include.quote }}""
```
igorkorenfeld, line 4: Extra quote mark here

igorkorenfeld (changes requested):
## Contact
- [ ] Sign up area has a `</Section>` showing up on the right hand side of the rendered section 
image

- [x] “Sign up for the 18F newsletter” let’s bump up to `xl`
- [ ] Sidebar missing borders above the headers. The patterns should follow the other sidebar designs, so this might need to be a separate issue / PR. 
Current site:
image

## WWD
### Cards
- [x] Lost the rounded corners

Some stuff I missed the first time around (included it on home but not here): 
- [x] Add 0.5 units of space between the agency name and the project title
- [ ] Chevron next to read more should be smaller and aligned with the x-height of the text

### Quote
- [x] The characters at the start of and end the quote are “prime-marks” (`&#8243;`), which are straighter,  instead of quote marks	(`&#8220;`  and `&#8221; `), which have a little bit of a curve. There is also an extra one mark at the end of the quote. Can we change to  (`&#8220; ` and `&#8221;`  / `&ldquo;` `&rdquo;`). I am wondering if that extra mark is what's causing them to be converted to prime marks instead of quote marks
- [ ] The Treasury logo is coming in a little pixelated, can we use an [SVG file](https://upload.wikimedia.org/wikipedia/commons/c/cb/Seal_of_the_United_States_Department_of_the_Treasury.svg)


## Footer
Looks good!

peterrowland (approved):
√

lauraGgit:
@igorkorenfeld and @peterrowland  thanks for the review.

@igorkorenfeld I made most of your changes, but the carets get fixed more globally in #3386 so holding off on those in this PR.

I also added the border to the sidenav but am holding off on more tweaks until a bit later to incorporate later changes. (too many prs!)

igorkorenfeld:
Looks good to me!

igorkorenfeld (approved):

PR #3411, opened 2021-03-25
Agency home list

lauraGgit:
# Pull request summary

Updates the agency list on the homepage 

(Optional) Closes #(issue_number).

- [x] update readme

- fix houndci feedback
- ensure tests pass
- federalist builds


lauraGgit:
@igorkorenfeld somehow the svgs look a little weird at this size?

Dahianna:
Oooh site is looking good! I do have a small suggestion. Should we add some space between the heading and the agencies? @igorkorenfeld

Dahianna:
Also we should update the agencies logos here too: https://federalist-c58f58dc-a215-4616-a54c-12b5eb011096.app.cloud.gov/preview/18f/18f.gsa.gov/agency-home-list/what-we-deliver/

lauraGgit:
@Dahianna on the what we deliver page, can we make that a separate ticket?

igorkorenfeld:
@lauraGgit  - I agree about the seals. I think it's that some of them have a lot of detail that's gettin compressed into a small area. At the same time we don't want the WWD page to be overwhelming with all of these agencies emblems (assuming we keep the same size for both). I suggest we increase the min-width size to 55px, which seems to improve visual, without it feeling cluttered. 

And yes @Dahianna agree, we should 2 units of space after the `h2` before the list of agencies

lauraGgit:
@igorkorenfeld @Dahianna thanks for the feedback, how is this?

Dahianna:
Looks good to me!

igorkorenfeld (approved):
Looks good!

PR #3432, opened 2021-04-01
Removes non-breaking space between words

brentryanjohnson:
This post has a non-breaking space (`&nbsp;`) between words, resulting in two spaces between "understanding" and "their". This patch removes the non-breaking space and replaces it with a single space.

Screenshot of blog post with two spaces between "understand" and "their" with inspect element showing a non-breaking space entity in the HTML output


Dahianna (approved):

Dahianna:
Thanks Ryan!

PR #3558, opened 2021-12-06
Make inclusion bot blog post accessible again

stvnrlly:
# Pull request summary

This is just a one-word metadata change that makes the inclusion bot blog post accessible on the site again.

PR #3562, opened 2021-12-15
Remove redirect, as well

stvnrlly:
# Pull request summary
This is a follow-on to #3558, since I missed the extra piece of metadata


geekygirlsarah (approved):
oops... that _might_ make a difference...

PR #3574, opened 2022-02-09
Remove head.html, add meta.html

Sgtpluck:
# Pull request summary

This PR:

- *Removes* `_includes/head.html` which does not seem to be called anywhere, and is not overwriting anything on the uswds--jekyll template.
-  *Adds* `_includes/meta.html` which will overwrite the `_includes/meta.html` from the [uswds-jekyll template](https://github.com/18F/uswds-jekyll/blob/main/_includes/meta.html). The basic uswds-jekyll metadata does not include social thumbnail details.
- Updates the image of my blog post to be the header image so it will be read in as the thumbnail image

Note: I am not 100% sure that the `head.html` file is not called anywhere, since I am not very familiar with Jekyll. It would be good if someone more familiar with it would give this a review!

- fix houndci feedback
- ensure tests pass
- federalist builds
- no new SNYK vulnerabilities are introduced

### DIFF
--- a/_includes/meta.html
+++ b/_includes/meta.html
@@ -4,22 +4,18 @@
 <!-- site.baseurl: {{ site.baseurl }} -->
 <!-- siteurl: {{ siteurl }} -->
 
-<head>
 
 <!-- Basic Page Needs
 ================================================== -->
-  <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


geekygirlsarah (approved):
Site works, looks the same as before, all the tests pass. I think it's good to merge!

Issue #3618, opened 2022-09-12
site.baseurl contains no value in circleCI


jduss4:
## Actual Behavior

Throughout the blog, people use `{{site.baseurl}}` in front of image paths. However, the value of `site.baseurl` is not set in `_config.yml`. Recently, during a CircleCI run, [html-proofer reported 4 broken links](https://app.circleci.com/pipelines/github/18F/18f.gsa.gov/1542/workflows/8407f711-8b1e-4e49-9386-0788fcfe0419/jobs/5735).  These links were broken because `site.baseurl` is empty.

![image](https://user-images.githubusercontent.com/2480492/189732524-4487b163-3502-4830-9bff-5bf5b994d300.png)
(image alt: four errors from `_site/2022/07/20/senior-executives-pt1/index.html` with links similar to `/2022/07/20/senior-executives-pt2`)

## Expected Behavior

The expected behavior is for html-proofer to be able to validate internal URLs in circleCI's environment while avoiding breaking functionality locally and in production.

Ideally, we would also document some of the additional `config` files, such as `_config-blog.yml` which have differing `site.baseurl` settings.

## Steps to reproduce the behavior

- [ ] Prepend an internal link to another page of the 18F website with `{{site.baseurl}}`
- [ ] Make a PR and confirm the html-proofer step of CircleCI fails

## This issue is done when

- [ ] Either `{{site.baseurl}}` is set appropriately or html-proofer running in a CICircle environment can check internal links through other means (perhaps a CLI flag or calling a different script from `.circleci/config.yml` )
- [ ] Internal links continue to work locally (both in docker and without) and in production
- [ ] Documentation has been added explaining how to use `site.baseurl`

Sgtpluck:
Fixed, as per #3619

PR #3619, opened 2022-09-12
changes internal links as temp workaround for circleci error

jduss4:
Alters a few links in recent blogposts to pass CircleCI broken links. This is a temporary fix, see https://github.com/18F/18f.gsa.gov/issues/3618 for more information about possible longer term solutions.


jduss4:
Good news, this fixed the original error.  Bad news, there's a different error:  
image

(alt text: error in 2014/11/17/taking-control-of-our-website-with-jekyll-and-webhooks that says iframe element requires a non-empty title attribute that identifies the frame)

PR #3621, opened 2022-10-03
Cannandev/18f guides combined

cannandev:
# 18F Jekyll Monorepo 

This prototype demonstrates pulling all guides into the 18f.gsa.gov GitHub repository. Ownership would fall to the same team monitoring the 18F website. It shares the same components required by the TTS Digital Council. It has new templates that accommodate different layouts and content formats. It also includes consistent front matter and data variables. This option requires redirects from existing urls to the new pages. For example, https://agile.18f.gov/ would redirect to https://18f.gsa.gov/agile. This change in URL structure means reconfiguring the search feature.

## Hypotheses tested in this prototype:

18F should have ownership of these guides
Separating code from content is attainable for static generated sites
GitHub UI is enough for content management
Breadcrumbs would enable users to get back to all guides
We need flexible page templates to accommodate different types of content 

## Feedback wanted:
How did you find the code when you wanted to make a change?
How would you get back to the other guides from the current guide? Are the other guides relevant to you?
Should the guides look like the 18F website?
Who do you think should own this guide? Should each guide have a separate owner?
Do you feel comfortable making updates to the guides? What would make you feel comfortable (WYSIWYG tools? Tutorials?)
Performance?
Dependency updates?
Editing a shared component?
Editing content?
Ease of contributing?
Automated scannability?

## Final pros/cons:


cannandev:
Moved to new repository https://github.com/18F/guides-and-methods/tree/branch-with-18f-site-combined

PR #3630, opened 2022-12-03
Hyphens are breaking tags

brentryanjohnson:
I noticed the format of the tags (with hyphens) in this post is preventing them from behaving like tags

This PR removes the hyphens so `open-source` and `open source` aggregate the same posts.


geekygirlsarah (approved):
Looks good to me, and makes sense about removing the dashes. Not sure if Laura/Julie need to approve this too?

geekygirlsarah:
I'm going to go ahead and merge this. It doesn't add/remove any tags and it matches the format of other tags on other pages. I think it fixes it while keeping stuff the same, so I think it's good.

PR #3643, opened 2023-01-12
Add featured card box-shadow on focus

### LINKED ISSUE #3642: Featured post cards: focus state
## Expected Behavior
On pages where we have featured posts (home page, how we work), in #3457 we made the whole card clickable, with some visual changes when you hover (increased shadow and header color change). We should make the same changes for :focus

## Actual Behavior
On focus, the cards do not visually change at all

## Steps to reproduce the behavior

- [ ] On the home page, click in the URL bar and tab through the links on the page
- [ ] When you get to the first card, there's no visible change (though it has focus, as indicated by the url that appears in bottom left of browser or if you press enter on your keyboard)

## This issue is done when

- [ ] On focus, the featured post card has the same shadow and changed header color as on hover

Sgtpluck:
# Pull request summary
This change:

- Moves box-shadow into css rather than via utility classes on the element directly
- Adds the box-shadow on a featured card when you tab to the element

(Optional) Closes #3642

- fix houndci feedback
- ensure tests pass
- federalist builds

### DIFF
--- a/_includes/card.html
+++ b/_includes/card.html
@@ -5,7 +5,7 @@
 
 <li class="usa-card tablet:grid-col-4">
   <a href="{{ card_link }}" aria-label="{{ project.title }}">
-    <div class="usa-card__container shadow-2 hover:shadow-5" tabindex="-1">
+    <div class="usa-card__container" tabindex="-1">
       
       <header class="usa-card__header">
         {% if project.agency %}
--- a/_sass/_components/card.scss
+++ b/_sass/_components/card.scss
@@ -5,10 +5,15 @@
   }
 
   a:focus {
-    div {
+
+    .usa-card__container {
       @include u-shadow(5);
       h3 {
         color: $color-dark;
+      }
+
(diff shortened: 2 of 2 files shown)


jstrothman:
Almost there! We could leave it as is, but I have 2 suggestions:
1. When you hover the shadow is only on the whole card (.usa-card__container) but when you focus, there's _also_ a shadow on the header image (.usa-card__image). (The shadow is set on `.usa-card a:focus div` and maybe should be on `.usa-card a:focus .usa-card__container` -- with that change you can pull .usa-card from the rule.)
2. New style request -- it's nagged at me that the hover/focus state is so subtle as to likely be unnoticeable some of the time. Could you underline the h3 on focus & on hover, e.g. something like: `a:focus .card-link-tagline {text-decoration:underline}; a:hover .card-link-tagline {text-decoration underline};`

jstrothman (approved):
Looks great!

PR #3659, opened 2023-02-28
Update dependencies; update build instructions

### LINKED ISSUE #3658: Update dependencies to run on modern hardware
Some of the transient dependencies are 5+ years old. As TTS begins to transition to a more mixed CPU environment (new Macs will be `ARM64` while Windows machines will continue to be `x86_64/amd64`), it is important that we update these dependencies so that they will work on multiple architectures.

mgwalker:
# Pull request summary

This PR updates a bunch of dependencies so they can be installed and run on modern Macs.

Closes #3658

- [x] update readme

- fix houndci feedback
- ensure tests pass
- federalist builds


mgwalker:
So this builds on my machine and in CircleCI (the test failure is because the `htmlproofer` command needs to be updated), but it's blowing up on cg Pages. No idea what the heck is going on with that. I'll look at the build logs again later to see if I can figure out what's up, but this is blocked until then.

mgwalker:
I recommend ignoring the html-proofer failures for this PR and create a separate issue to fix them. They are all errors because of anchor tags with empty `href` attributes.

PR #3695, opened 2023-05-11
Jskinne3 fix footer order

jskinne3:
# Pull request summary
Move the closing footer tag to the bottom of the footer. The result of this is to:
1. re-order footer content (to what I believe is the correct order) when the browser window is wide
2. correct tab navigation order of the footer, per this ticket: https://github.com/18F/TLC-crew/issues/200


aduth (approved):
I think a root issue is that there's [a `footer` style changing the order](https://github.com/18F/uswds-jekyll/blob/ddedf992becc3b75e30f808b3231df403cd90cc7/_sass/uswds/components/_footer.scss#L3). Nullifying that would also fix the issue, and it'd probably be good to remove and avoid `order` altogether at the source if possible, given the accessibility implications.

Design system [guidance](https://designsystem.digital.gov/components/identifier/) seems to consider identifier as separate from the site's main footer, so combining the two may be unexpected, but I don't know if it'd have much impact for better or worse in practice.

jasnakai (approved):
LGTM!

jskinne3:
> I think a root issue is that there's [a `footer` style changing the order](https://github.com/18F/uswds-jekyll/blob/ddedf992becc3b75e30f808b3231df403cd90cc7/_sass/uswds/components/_footer.scss#L3). Nullifying that would also fix the issue, and it'd probably be good to remove and avoid `order` altogether at the source if possible, given the accessibility implications.
> 
> Design system [guidance](https://designsystem.digital.gov/components/identifier/) seems to consider identifier as separate from the site's main footer, so combining the two may be unexpected, but I don't know if it'd have much impact for better or worse in practice.

Thanks for the detailed feedback! After experimenting and looking at other USWDS repos I decided to keep my expansion of the `footer`, and also per your advice to unset the order — but just within the project, rather than at the source in the [uswds-jekyll](https://github.com/18F/uswds-jekyll) gem

PR #3697, opened 2023-05-17
Remove the FOIA reading room page, since it has never been maintained

mgwalker:
# Pull request summary

The 18f.gsa.gov site has a [FOIA reading room page](https://18f.gsa.gov/foia-reading-room/) that has not been updated since 2017. We've responded to FOIA requests since then, but we didn't update the page. Since we don't update the page, we should just take it down because that's more honest. The existence of the page implies that it's maintained.


jskinne3 (approved):
This looks great! I strongly agree that removing never-used documentation is more honest.

I fixed CI checks on the main branch. If you want CI to pass for this PR, rebase or merge main.

geekygirlsarah (approved):
Build passes since merging in `main`, so I think it's good to merge now.

PR #3701, opened 2023-06-07
Tlc 237 how we work accessibility bugs

mel-keene-gsa:
# Remediate 18F How We Work accessibility bugs

https://github.com/18F/TLC-crew/issues/237

### DIFF
--- a/pages/how-we-work.md
+++ b/pages/how-we-work.md
@@ -16,7 +16,6 @@ content_focus: false
 
 </div>
 
-* * *
 
 {% capture pa-1 %}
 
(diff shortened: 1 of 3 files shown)


pages/how-we-work.md
jasnakai, line 1: Did these `<hr>`s get replaced with CSS?

mel-keene-gsa: does this work? as far as i can tell the <hr> is using the styles in https://github.com/18F/18f.gsa.gov/blob/tlc-237-how-we-work-accessibility-bugs/_sass/_templates/how-we-work.scss 

not sure how to get the aria-hidden into the css as opposed to the element itself...? would you be up for pairing on that sometime?

jskinne3 (approved):
LGTM

PR #3714, opened 2023-07-18
Redirects for What we deliver and How we work

michelle-rago:
Renaming page and attempting a redirect from the old page


michelle-rago:
@AnitaYCheng Are you okay with me merging this redirect branch into the tlc branch we've been working in? Note that I didn't update every instance of how-we-work and what-we-deliver (including tags). I thought that could be a cleanup activity if needed.

PR #3811, opened 2024-03-04
Add hyphen to user-centered design tag

juliaklindpaintner:
Adds a hyphen in "user-centered design" in the tag that pulls blog posts on the website together. Currently two blog post have the tags listed without hyphens, which prevents all posts from being shown.


neilmb (approved):
Looks like a great fix.

Issue #3752, opened 2023-09-07
Pa11y runs are too slow


alexsobledotgov:
## Problem

* Right now the CI takes a very long time to run: ~15-20 minutes for each commit. 
* The pa11y CI step is the longest and takes about ~15 minutes to run.
* This CI step scans every single page of the 18F website every time it runs. It's re-scanning the entire website unnecessarily, over and over and over.

As of 4/12/24: There's been a lot of activity on this ticket! We're moving it forward with the following acceptance criteria:

## Acceptance criteria
- [x] Adjust the pa11y-limited [npm script](https://github.com/18F/18f.gsa.gov/blob/main/package.json#L17) to scan appropriate pages for each Pull Request. (Will be up to the developer to determine pa11y-ci's capabilities here in terms of what sub-set of pages can be reasonably scanned.)
- [x] Use pa11y-limited script for pull request CI pipeline
- [x] Find a way to run the [complete scan](https://github.com/18F/18f.gsa.gov/blob/main/package.json#L16) on some decided-upon schedule (once a week, once a month, etc) outside of the git workflow, and notify a maintainer of any errors. Note: as an example, Tock also runs [a job on a daily schedule](https://github.com/18F/tock/blob/main/.circleci/config.yml#L178-L180). Feel free to reuse the Tock configuration. As for notifications, we do an audit every week which lets us catch CI errors, but that probably won't work for TLC.
- [x] Document how the notifications of broken scheduled runs work — who gets notified and how.

echappen:
Path forward for this ticket may be determined by outcome of [TLC-crew #474](https://github.com/18F/TLC-crew/pull/474)

echappen:
Issue that blocks this work is [here](https://github.com/18F/TLC-crew/issues/501)!

juliaklindpaintner:
Thanks @echappen! Can you link to the ADR as well? And then I think we decided we could move this one to done as the ticket names a specific solution. Let me know if you disagree or think the content of the ticket should be captured in the other ticket.

echappen:
To recap the story of TLC’s pa11y work that came out of this issue:

1. This issue was incorporated into[ this TLC issue](https://github.com/18F/TLC-crew/issues/472) to explore any and all ways to make pa11y faster (without proposing a solution)
2. Here is the [old proposed ADR for looking into above issue](https://github.com/18F/TLC-crew/pull/474)
[...]

PR #3829, opened 2024-05-02
Reduce pa11y runtimes

caleywoods:
# Pull request summary
Attempts to address #3752. 

We're currently running pa11y scans using the sitemap which scans all of the URLs on the site which results in slowness. These changes use a two pronged approach to reducing the pa11y runtimes. 

## The Jekyll plugin
The first part is a Jekyll plugin that uses Jekyll hooks (`:documents`, `:pages`, and `:site` currently) to output a file named `pa11y_targets` to the filesystem during the build process. The plugin has the following behavior:
* When changes are made to a post or to any file in a [user defined collection](https://github.com/18F/18f.gsa.gov/blob/main/_config.yml#L36), those will be picked up by the `:documents` Jekyll hook and be included in the list of files to be scanned by pa11y. This takes care of the majority of file changes
* If a change is made to a file in the `pages/` directory, the `:pages` Jekyll hook captures those to be included in the list of files to be scanned by pa11y (I'm not sure why `:documents` doesn't get these)
* If a layout is changed the plugin determines what files use that layout and includes all of them to be scanned by pa11y
* If a change is made to a "site-wide" file in `assets`, `_includes`, or `_sass`, the plugin does a sampling of pages across the site. These pages are: the homepage, three files that use each layout from `_config.yml`'s `defaults`, as well as the first, middle, and last blog archive pages (`/blog/page/:num`)
* All files that are considered changed are written to the `pa11y_targets` file and each path is separated by a newline

There's RSpec tests to go along with this plugin.

## The CircleCi config
The second part is a tweak to `.circleci/config.yml` when the Jekyll build step runs. This change does two things:
* It makes the contents of the `pa11y_targets` file be base64 encoded and exported into an environment variable called `PA11Y_TARGETS` in [CircleCI's `$BASH_ENV` area](https://circleci.com/docs/env-vars/#parameters-and-bash-environment) which is stateful between job steps
* Two job steps later when pa11y is initiated, it reads the list of files to scan out of the previously stored `PA11Y_TARGETS` environment variable

## Results
Without these changes, the full build takes around 25 minutes in CircleCI with the pa11y scan representing the bulk of that time taking around 20-21 minutes. Testing these changes showed CircleCI full builds in under 4 minutes with the pa11y scan portion being reduced to < 10 seconds in most cases. These are big gains for the common scenario of single file changes when someone is creating or editing a blog post.

 - [CircleCi current build example, 25m 35s runtime](https://app.circleci.com/pipelines/github/18F/18f.gsa.gov/2414/workflows/42fa5cdb-4a4c-4b8a-9dcc-9894e8aeee53/jobs/6622)
 - [CircleCI `_layouts/styled-container.html` was changed, 3m 49s runtime](https://app.circleci.com/pipelines/github/18F/18f.gsa.gov/2431/workflows/65035aeb-7e1e-4a54-95dd-9289f788fe01/jobs/6639) (Note: this scan was prior to the `:pages` hook being added, `_layouts/styled-container.html` applies to three pages in the `pages/` directory but adding three more files to the pa11y scan would not have caused a significant increase in the pa11y scan time)
 
 This could still have some long running build times/pa11y scans depending on what files are touched. An example would be `_layouts/post.html`, when I checked locally, changing that particular layout resulted in pa11y scanning a little over 500 files.


caleywoods:
Tagging @beechnut - I know you're not on TLC crew quite yet but you did a lot of great work on this and wanted to make sure you had a chance to look at it.

_plugins/pa11y.rb
```diff
@@ -0,0 +1,177 @@
...
+# @todo Is there a way to get these more dynamically?
+SITEWIDE_FOLDERS = ["assets", "_includes", "_sass"]
+
+# If files have changed in any of the folders that trigger a site-wide scan,
+# sample folders across the site and add them to the list of files to check
+# with pa11y for a given Pull Request.
+# @todo Is there a way to unnest this code? Jekyll hooks don't allow
+#   early returns.
+Jekyll::Hooks.register :site, :post_render do |site|
+  # https://rubular.com/r/nfFL9P69KHSBoY
```
beechnut, line 168: clarification: What's the context for this comment? And/or can it be removed?

caleywoods: The comment is providing an example of the generated RegEx against an imagined list of changed files to highlight what the regular expression is doing in case someone needs to change the code who is unfamiliar with regex.

_plugins/pa11y.rb
```diff
@@ -0,0 +1,177 @@
...
+def check_file_for_changes(file)
+  document = Document.new(
+    file.relative_path,
+    file.data["layout"],
+    DIFFER
+  )
+
+  if document.to_scan?
+    File.open(PA11Y_TARGET_FILE, 'a') { |f|
+      f.write(file.destination(NO_ROOT_PATH) + "\n")
```
beechnut, line 80: suggestion (non-blocking): Re-reading this, I'm finding it confusing to have the narrative go: `document.to_scan?` followed by `file.destination`.

I think a clearer refactoring would involving write `document.destination` to the file instead, where `document.destination` is `file.destination(NO_ROOT_PATH)`.

This is not so important that it should block merging, but rather an optional suggestion.

caleywoods:
I also wonder if it's detrimental to never run a full pa11y scan. Should we consider adding a scheduled GitHub action to run a full scan maybe every two weeks?

_plugins/pa11y.rb
```diff
@@ -0,0 +1,177 @@
...
+end
+
+# Jekyll::Document#destination requires a root path parameter
+# This constant replaces a mysterious empty string when
+# calling #destination.
+NO_ROOT_PATH = "".freeze
+PA11Y_TARGET_FILE = ENV.fetch('PA11Y_TARGET_FILE') { 'pa11y_targets' }
+DIFFER = ENV.fetch("CI", false) ? CommitDiffer : Differ
+
+def check_file_for_changes(file)
```
beechnut, line 71: suggestion: This function is at the top level, so when it's `required` into the test file, we'll have a free-floating function. **I'd suggest, at minimum, organizing this into a module,** probably a new one.

suggestions (non-blocking): My dream is to have this function be tested, and that would be easier if the pa11y target file and differ constants were both able to be dependency-injected.

spec/_plugins/pa11y_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    context "with site files and permalinks" do
+      let(:site_files) { ["_site/blog/index.html", "_site/blog/page/2/index.html"] }
+      let(:permalinks) { ["blog/page/:num/", "blog/"] }
+      let(:collector) { SiteSampler.new({}, site_files, permalinks) }
+      it "lists pages" do
+        expect(collector.pages.count).to eq(2)
+      end
+    end
+    
+    context "ignores PDF files" do
```
beechnut, line 90: This `context` is describing something better described with `it`.

I'd suggest changing this to:

```
context "with HTML and PDF files" do
  it "ignores PDF files"
```

beechnut:
@caleywoods Yes, part of the scope of work is to add a nightly full scan. We shouldn't remove full scans from the pipeline without replacing them beforehand or in the same PR.

_plugins/pa11y.rb
caleywoods: Yep - I will add that to the list of changes, thanks!

cantsin (approved):
LGTM!

_plugins/pa11y.rb
caleywoods: I didn't quite understand this one, were you thinking it should update `Document` to have the files destination method so it can replace `file.destination()`?

caleywoods:
> @caleywoods Yes, part of the scope of work is to add a nightly full scan. We shouldn't remove full scans from the pipeline without replacing them beforehand or in the same PR.

I've opened #3832 to track this and I plan on working on it this week.

_plugins/pa11y.rb
beechnut: Ideally, it reads cleaner. But not a big enough deal to not merge.

PR #3834, opened 2024-05-16
[WIP] Nightly full pa11y scan

### LINKED ISSUE #3832: Add nightly pa11y scan
## Feature suggestion

Create a GitHub/CircleCI action that runs a nightly pa11y scan. If this scan results in pa11y finding issues, it should create a GitHub issue with the details.

caleywoods:
# Pull request summary
#3829 Changed how pa11y scans are handled in CI so that instead of full site scans using the sitemap it only scans the files that were changed. Part of that scope of work was to create a scheduled full site scan with pa11y since that now never occurs otherwise.

There are three parts to completing this work, only the first step is complete currently:

1. Modify the CircleCI config so that it supports both a scheduled full scan and the normal "run a changed files scan when a commit to a branch that has a pull request open" scan [ ✅ ]
2. Create a GitHub action that listens for the `repository_dispatch` event and creates a GitHub issue from the pa11y failures, those pa11y failures are provided in the event that will be sent to GitHub from our CircleCI step listed in step 3 [ ✅ ]
3. Make the scheduled nightly scan create a GitHub issue if there's a pa11y failure. The current plan is to add a step to the CircleCI config that uses cURL to send an event through the GitHub API to the `18f/18f.gsa.gov` repo that triggers the `repository_dispatch` GitHub action [ ✅ ]

You can see a successful run of the scheduled "nightly" execution [here on CircleCI](https://app.circleci.com/pipelines/github/18F/18f.gsa.gov/2467/workflows/a7c70471-5247-4bd7-b165-97a8521851ec/jobs/6669).

Attempts to close #3832 

Current status: tweaking the logic in our GitHub action responsible for creating the GH issue when a scan fails so that it tries to handle issue creation intelligently. The goal is to have it not open a new issue if there's an issue already open and the scan result between the two runs was identical.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
       # which satisfies this condition
       - when:
           condition:
-            equal: [ 'scheduled_pipeline', << pipeline.trigger_source >> ]
+            equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
           steps:
             - run:
                 name: Run full pa11y scan
@@ -70,7 +70,7 @@ jobs:
       # it makes a webhook POST to CircleCI that satisfies this condition
       - when:
           condition:
-            equal: [ 'webhook', << pipeline.trigger_source >> ]
+            equal: [ webhook, << pipeline.trigger_source >> ]
           steps:
             - run:
                 name: Run pa11y scan on changed files
@@ -87,7 +87,7 @@ workflows:
     triggers:
       - schedule:
           # Run nightly at 10am UTC / 5am ET
-          cron: "15 * * * *"
+          cron: "25 * * * *"
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,7 +117,7 @@ workflows:
     triggers:
       - schedule:
           # Run nightly at 10am UTC / 5am ET
-          cron: "55 9 * * *"
+          cron: "45 * * * *"
           filters:
             branches:
               only:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -42,23 +41,59 @@ jobs:
...
+          name: Jekyll build and store pa11y_targets
+          # We export the changed file list in pa11y_targets source
+          # that we can make pa11y-ci scan only those changed files
+          # in the "Run pa11y scan on changed files" step if it's
+          # needed, a full site scan does not use the pa11y_targets file
           command: |
             bundle exec jekyll build
-            echo "export PA11Y_TARGETS=$(cat pa11y_targets | base64 --wrap=0)" >> "$BASH_ENV"
-            source "$BASH_ENV"
+            if test -f pa11y_targets; then
```
caleywoods, line 52: todo: I need to add an else to this check, if no files were changed that would cause the pa11y_targets file to be written to the file system we should still push some value into the environment variable that we can apply conditional logic to for the Pa11y scan step which could be skipped all together. I'm thinking of a change to a file(s) that wouldn't create the pa11y_targets file is things like `.eleventy.js` or dependabot updating `package.json`.

caleywoods: After some more review and testing, this doesn't need an else, I mistakenly had the "Do a pa11y scan on the changed files" logic placed inside the `if [[ -z "${PA11Y_TARGETS}" ]]; then` check when that if statement is actually checking for a zero length result which happens when the env var is empty or unset.

caleywoods:
Closing this - it will be re-opened soon based on the replatforming work.

PR #3871, opened 2024-06-12
Replatform: fix checks

cantsin:
Fixing check failures as we go along.

- [Cloud.gov pages for 18f.gsa.gov](https://pages.cloud.gov/sites/3/builds) now builds properly, allowing us to have previews
- Github workflows are [temporarily disabled](https://docs.github.com/en/actions/using-workflows/disabling-and-enabling-a-workflow) so we aren't indundated with emails 
- Removed CircleCI checks that have to do with ruby/bundler
- Disabled pa11y checks until we are in a place to get it to work again with @caleywoods' newer work
- Added build, lint, html validation, and link testing to the build steps so that checks are immediately useful and reflective of build quality in the PR

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,11 +16,6 @@ jobs:
             sudo locale-gen en en_US en_US.UTF-8
             sudo dpkg-reconfigure locales
       - checkout
-      - save_cache:
-          paths:
-            - vendor/bundle
-          key: v2-dependencies-bundler-{{ checksum "Gemfile.lock" }}
-
       - restore_cache:
           keys:
           - v2-dependencies-npm-{{ checksum "package-lock.json" }}
@@ -31,9 +26,6 @@ jobs:
           paths:
             - node_modules
           key: v2-dependencies-npm-{{ checksum "package-lock.json" }}
-      - run:
-          name: Check Ruby style
-          command: bundle exec standardrb
       - run:
           name: Check JavaScript style
           command: npx standard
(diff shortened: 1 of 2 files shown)


beechnut:
@cantsin Thanks, these checks were annoying me as well.

I want to propose something, let me know what you think.

There are a few steps that should all be working, at this point in the migration:
- npm run lint
- npm run build
- npm run test:html-validation
- npm run test:links

Do you think it would be within scope to add back these steps? Or should we wait and add these later?

cantsin:
Great idea, we want these checks to be useful as well. Adding!

.circleci/config.yml
```diff
@@ -41,28 +26,28 @@ jobs:
...
-          name: Check Ruby style
-          command: bundle exec standardrb
-      - run:
-          name: Check JavaScript style
-          command: npx standard
-      - run: npm run uswds-build
+      # - run:
+      #     name: Check JavaScript style
+      #     command: npx standard
+      - build: npm run build
```
beechnut, line 32: suggestion (non-blocking, we can do it another time): Running `lint` before `build` would fail faster, and `lint` doesn't depend on `build`.

.circleci/config.yml
```diff
@@ -41,28 +26,28 @@ jobs:
...
+      - lint: npm run lint
+      - verify:
```
beechnut, line 34: Interesting — I wouldn't have thought of grouping these!

beechnut (approved):
Looks good to me! Just one non-blocking suggestion we can do another time.

When you merge, please select "Squash and Merge" so these come over as one commit.

PR #3874, opened 2024-06-13
Adjust navigation and layouts

### LINKED ISSUE #3866: Migrate navigation
There are a number of includes/layouts that handle navigation, and we should migrate those as-needed.

A few of the filters in the Ruby files, like `matches_url` and `hash_link`, support the navigation and would need to come over.

```[tasklist]
### Tasks
- [ ] Assess whether nav layouts need to come over
- [ ] Migrate or delete utility filters
```

beechnut:
# Pull request summary
Removes unused nav partials, and makes small changes to layouts to ensure related posts and footers are rendering correctly.

Closes #3866


jasnakai (approved):
LGTM!

PR #3876, opened 2024-06-13
Add GH action for creating issues when a nightly pa11y scan fails

caleywoods:
# Pull request summary
This PR adds a new GitHub action that kicks off when a [`repository_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) webhook event is received. This webhook is currently only called from CircleCI if there are failure(s) encountered during the scheduled nightly pa11y scan.

I'm separating this workflow file/action out into a new branch/PR because the action needs to exist on the default branch (`main`) in order for me to test my webhook call.


cantsin (approved):
LGTM

PR #3949, opened 2024-06-19
Add pa11y scanner

### LINKED ISSUE #3875: Migrate pa11y plugin
Migrate pa11y plugin so builds report the files to scan for the per-PR pa11y run

```[tasklist]
### Tasks
- [x] Per-page scan
- [x] Sitewide scan
```

beechnut:
Previously, we did accessibility scanning for every file in every push / build / PR, which took > 25 mins each run.

Now, we more carefully curate the files that need to be scanned, to limit the scan time.

This PR:

- Adds a per-page pa11y scan, which lists pages that have changed or whose layouts have changed.
- Adds a site-wide pa11y scan: if files that affect the whole site change, take a sample of files from across the site to check
- Incidentally deletes an unused layout
- Pins the glob package to 8, because 9 and higher breaks something with the SVG Sprite package.

Commentary:

This commit gets us to the "Make it work" stage, not yet addressing "Make it right" or "Make it fast".

The code also reads a little weird, and I'm not sure a conditional "eleventy.after" event is the right design for this. I think one more refactor here would be wise.

Closes #3875


lib/pa11y.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+  return result
+}
+
+function doesIntersect(setA, setB) {
+  return intersection(setA, setB).size > 0
+}
+
+function didSitewideFilesChange() {
+  const result = doesIntersect(changedFileSet(), sitewideFileSet())
```
caleywoods, line 66: Non-blocking suggestion here, I might rename `result` to `filesChanged`.

lib/pa11y.js
```diff
@@ -0,0 +1,154 @@
...
+
+  results.forEach((result) => {
+    const doc = new Document(result, dir, changedFiles);
+    if (doc.shouldScan()) { filesToWrite.push(result.outputPath) }
+  });
+
+  return filesToWrite
+}
+
+async function pa11yScan ({ dir, results }) {
```
caleywoods, line 128: Non blocking question/suggestion: `results` sounds like it's an array of pages/files/documents sent from `eleventy.after`, if that's correct then maybe renaming `results` to have a little more context could help readability? If so I think the same renaming should occur where `results` is passed to the `pa11yScan` function.

caleywoods:
I added some non blocking suggestions related to readability, this feels like a great start Matt!

caleywoods (approved):

PR #3959, opened 2024-07-01
Better local dev server

### LINKED ISSUE #3819: Serve command runs in background
## Expected Behavior

When I run `./serve` and signal an interrupt (`ctrl+C`), the `serve` command should stop running and the site should stop working in the browser.

## Actual Behavior

When I run `./serve` and signal an interrupt (`ctrl+C`), `serve` continues running in the background and the local site continues to work.

## Steps to reproduce the behavior

- [ ] Follow local installation instructions and run `./serve`
- [ ] Visit the site (it should be running)
- [ ] Type `ctrl+C` to interrupt
- [ ] Refresh the tab with the site (it's still running, but shouldn't be)

## This issue is done when

- [ ] `serve` shuts down when given an interrupt signal (`ctrl+C`)

beechnut:
Something is wrong with the `./serve` command — that `&` detaches a process, and I'm often running into port conflicts that have to be shut down with `killport`.

There's a better way.

This commit:

- Adds a Foreman-based local dev server:
  - Adds a Procfile for building assets and running the server
  - Adds instructions to the README for using this
  - Corrects references to localhost in the README

You can still use the `./serve` commands if they work for you.

Closes #3819

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+assets: npm run uswds-build && npm run uswds-watch
+web: rm -rf _site/ && LC_ALL="en_US.UTF-8" bundle exec jekyll serve --trace --incremental --host 0.0.0.0 --baseurl /site
--- a/README.md
+++ b/README.md
@@ -8,10 +8,7 @@ This repository houses the 18F website. We use the [U.S. Web Design System](http
 
 ## Local development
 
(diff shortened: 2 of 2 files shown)


caleywoods (approved):
Looks good! Worked perfectly for me locally.

Issue #3971, opened 2024-07-09
Replace Jekyll inline styles


beechnut:
Jekyll allows you to do inline styles, but 11ty doesn't. Find them by searching for `{:` in .md files.

All of them seem to be related to styling images or their captions. Instead of using inline styles, let's replace them with 

Fortunately, there are only 7 of these across the site.

```[tasklist]
### Acceptance criteria
- [ ] There are no instances of Jekyll-style inline styles `{:` across the site
- [ ] The blog posts that use to have them render about the same
```

PR #3974, opened 2024-07-11
Replacing Jekyll styles with 11ty-friendly markdown

sam-couch-18f:
# Pull request summary
Replaces Jekyll inline styles with 11ty-friendly HTML classes, using [markdown-it](https://github.com/markdown-it/markdown-it) (which is already included in the project). 

Related to TLC ticket https://github.com/18F/18f.gsa.gov/issues/3971


beechnut:
@sam-couch-18f OMG was the fix here just removing the leading colon? That was it?!

beechnut (approved):

PR #3975, opened 2024-07-11
Revert to using posts.json

beechnut:
Revert to using posts.json

It's a bit of a mystery to me why posts.liquid and posts.json differ when using the same syntax, but it appears the site doesn't build most posts when it's posts.liquid, but it works fine when it's posts.json. Maybe it's something to do with how the data pipeline works.

So, we'll leave it as JSON.


beechnut:
Self-merging to correct the previous self-merge that introduced this. Apologies, everyone.

PR #4006, opened 2024-07-17
Update heading levels for oracle blog post

igorkorenfeld:
# Pull request summary
This PR shifts the headings in the databases blog post down one level, so that there is only a single `h1` headings on the page. Our blog uses the post title as an `h1`, so I've shifted every other heading by 1 (`h1`-> `h2`, `h2`->`h3`, etc...). The exception to this this was the last two headings for the Conclusion and Special Thanks, which seemed like they were not part of the "Tutorial Part 2" section, but if that's incorrect happy to adjust. 

- [ ] Update readme
- [ ] For content changes beyond typos, add Ron Bronson as a reviewer

### DIFF
--- a/_posts/2024-07-17-working-with-oracle-databases-in-open-source-projects.md
+++ b/_posts/2024-07-17-working-with-oracle-databases-in-open-source-projects.md
@@ -16,7 +16,7 @@ excerpt: >
 
 _Reference to any non-federal entity does not constitute or imply its endorsement, recommendation, or favoring by GSA._
 
-# Introduction
+## Introduction
 
 During a recent 18F project, we built an open-source prototype with a twist: all of the data for our project was locked in a closed-source Oracle database, with no way to export it to an open format.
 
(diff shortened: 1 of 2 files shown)


beechnut (changes requested):
First of all, thanks for noticing this. Since this is a site-wide rule that's not listed anywhere as far as I can tell, we should build this check in to the site itself — I've added this to the [18F replatforming post-migration tasks]().

I don't like the way the h4's look in the preview build — when scanning, they almost read as body text, and don't break the content up visually very well.

I think I'd significantly prefer knocking just the h1s down to h2s and leaving the rest as-is. It may not be as semantically correct, but having them stay larger better fulfills their purpose as separators.

igorkorenfeld:
Great, I've noticed the issue on a number of blog posts, so that'll be a good upgrade.  

A possible alternative which would keep the semantic structure would be to style the `h4` and `h5` a bit larger in the blog posts?

cantsin:
We can have our cake and eat it too -- we can have semantically correct headers and sizes to match, but maybe this is a separate (a11y related?) ticket.

igorkorenfeld:
@beechnut I've update the styling for the headings to give the `h4`s and `h5`s a bit more prominence. The added benefit is that this better matches our intended design from last year too :) 

Let me know how this looks to you

beechnut (approved):

beechnut:
Wow, this looks great! Love the new styling! Approved!

PR #4013, opened 2024-07-18
Refactor configuration files

beechnut:
Problem: The configuration files have been disorganized, causing confusion about where to look for what.

This commit follows the [Handbook's scheme of having one main plugin that inherits everything](https://github.com/18F/handbook/blob/main/config/HandbookPlugin.js), and using that plugin as [one of few things required in .eleventy.js](https://github.com/18F/handbook/blob/main/.eleventy.js#L14).

This keeps the main configuration files small, and all of the rest of the configuration in nicely organized folders.


cantsin (approved):
yesssssssssssss!

beechnut:
Wow that feels good

PR #4020, opened 2024-07-23
Related posts

### LINKED ISSUE #3856: Migrate related posts
The 18F website recommends "related posts" based on posts with similar tags and by the same author.

The Jekyll site used a generator to add "related posts" to a page based on categories/tags/authors — basically, a post was more related if it shared tags, authors, and categories.

It's not a requirement that the generator code is translated exactly from the Jekyll site, but it should serve as a good starting point for an algorithm. I don't recommend trying to understand it as-written — rather, refactor it to understand it.

The scope of work for this ticket is to write a good-enough shortcode or filter that takes a post and optional limit, and returns the most related posts by tag and author (we don't use categories). If several posts have equal or equivalent relatedness scores, sort by newest-first.

```[tasklist]
### Acceptance smoke tests
- [ ] The ["Senior executive"](https://18f.gsa.gov/2022/07/20/senior-executives-pt1/) series posts all have each other as related posts
- [ ] The ["engineering sandwich" post](https://18f.gsa.gov/2022/09/06/18f-tech-sandwich/) has all technical / open-source related posts.
```

beechnut:
# Pull request summary

Migrates "related posts" functionality, so that blog posts show three relevant other posts in the footer. This adds a negligible amount of time to the build, maybe 3–5 seconds on a 2019 Intel Mac.

For the PR reviewer, the general process is:

- Compare all posts to all other posts, calculating their similarity scores. We avoid calculating a post combination's score twice.
  - The similarity score is: +1 for common author, +1 for common tag, 2x multiplier if a post shares at least 1 tag and 1 author
  - We calculate similarities starting with the most recent first, so, with a set of posts where all the similarity scores are equal, the related posts that appear should generally be the more recent ones.
- Then, build a file `.cache/related-posts-{timestamp}.json` that stores all the posts and the essential data for rendering their related posts
- In the template, use the cache file to fetch the related posts and render them
  - In the template, we look up the post's permalink URL from the post's input path, and use that to link the post. If you're wondering, "why don't we store the post's permalink URL as the key instead?", it's because we don't have access to it, and Eleventy is built in such a way that I'm not sure there's even a way to access it. (v 3.0 pre release has a filter to solve this problem, but we're on 2.0)

The cache is regenerated in response to posts changing, that is, the file only rebuilds when there's any changed files in `content/posts/*.md`. In an ideal world, we'd check only for new posts and for posts whose authors or tags have changed, but that would be a lot of coding time for minimal — if any — build time saved.

The cache files should be checked into version control, because unless you're changing a post, you should be using a stable set of related posts. The `Cache` class deletes old cache files, so only the most recent cache file will appear in `.cache/`.

Closes #3856


config/data/index.js
```diff
@@ -0,0 +1,8 @@
+// Add data sources here
+const dataPlugin = (eleventyConfig) => { /* eslint-disable-line no-unused-vars */
```
cantsin, line 2: Do we need this file?

cantsin (approved):
LGTM, one question about an empty plugin but looks great!

config/data/index.js
beechnut: Quite a story here: I created this file thinking I'd connect the related posts as a data source (which did not work), then I went to delete it, then I realized that someone is bound to eventually add data to 11ty config and wonder where to put it, so I was like "I might as well leave this for them"

So my answer would be "no but actually yes", but that's a loosely held opinion.

PR #4045, opened 2024-08-08
Add "Speaker bureau" page

AnitaYCheng:
# Pull request summary
Add a new `/speaker-bureau` page and a link to it. Includes 2 images.


beechnut:
@AnitaYCheng Before you do much more work on this, just a heads-up that we're probably days away from switching over to the replatformed site.

@cantsin Thoughts on whether we should do a code freeze on `main` / ask that changes be made against `replatform-main` starting now?

cantsin:
@AnitaYCheng What is the timeline/urgency of this change? If it's not time sensitive, then I concur with Matt; we should freeze any changes for now, yes. We are close to switching over.

AnitaYCheng:
> @AnitaYCheng What is the timeline/urgency of this change? If it's not time sensitive, then I concur with Matt; we should freeze any changes for now, yes. We are close to switching over.

Not urgent! Totally cool with ditching this branch and tackling it on the replatformed site.

Issue #3864, opened 2024-06-10
[EPIC] Post-migration tasks


beechnut:
```[tasklist]
### Tasks
- [ ] #3862 
- [ ] Tags are inconsistently capitalized (see sidebar in /blog vs /tags)
```

bpdesigns:
@beechnut please turn these tasks into individual tickets with acceptance criteria so we can prioritize these. Thank you!

beechnut:
@bpdesigns I'm of the opinion that these shouldn't be addressed or prioritized until the migration is fully complete. Thoughts?

bpdesigns:
makes sense to me. I'll add them to the icebox for now. Thanks!

beechnut:
@amandacostello I moved this into "Candidates for prioritization" now that the replatforming is done.

PR #4057, opened 2024-08-13
Manually correct social media external links

### LINKED ISSUE #4053: replatform: external links are not marked appropriately
We are missing `usa-link--external` especially for social media links. This is a weird one because my local copy of 18f.gsa.gov builds without, so I didn't catch this the first time around. 

Example: 

http://localhost:8080/2019/05/29/you-might-not-be-as-agile-as-you-think-you-are/ 

20240812_10h37m00s_grim

versus

https://18f.gsa.gov/2019/05/29/you-might-not-be-as-agile-as-you-think-you-are/

20240812_10h37m05s_grim

Also missing these for the footer, e.g., in the "Social" section.

beechnut:
Solution options here were:

1. spend a long time figuring out why the Markdown renderer that's supposed to handle links isn't seeing the "18F on {platform}" links in the social media sidebar, or
2. manually mark them external

I opted for 2 just to get the migration done for now.

I've listed a task in #3864 to handle this more systematically — our renderers not in good condition to be maintainable in the long run, so this is pennies on top of an already hefty technical debt.

Fixes #4053

### DIFF
--- a/data/social_media.yml
+++ b/data/social_media.yml
@@ -1,18 +1,27 @@
 -
   platform: GitHub
   link: https://github.com/18F
+  html:
+    class: usa-link--external
+    rel:   noreferrer
   image:
     light: assets/img/social-icons/svg/github-lightest.svg
     dark: assets/img/social-icons/svg/github-darkest.svg
 -
   platform: Twitter
   link: https://twitter.com/18F
+  html:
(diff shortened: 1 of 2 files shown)


cantsin (approved):
LGTM

PR #4058, opened 2024-08-13
Fixing things I am finding as I go through

cantsin:
# Fixes

- Tags have double // in the url
- "Contact" template front matter fixes
- Footer has explicit external link
  - Added for "Report a bug" as well
- Added `subnav.html` back. 
  - **Note** I removed `hash_link` tag and removed `{% assign is_matching = page.url | matches_url: item.permalink %}` but I don't think this broke anything. I believe `is_matching` was broken on the original site too -- it doesn't seem to work.
  - To test subnav, visit `/join`, `/about`, or `/partnership-principles`
-  `hide_footer_rule`
- Remove `_posts/2015-08-12-18f-consulting-is-hiring.md` as this is an unpublished post 
- `usa-current` for blog posts
- missing blog image
- fix pagination offset by 1
- `/guides` paragraph styling at the bottom
- bring back "Recent posts"


content/pages/tags.html
```diff
@@ -8,7 +8,7 @@
 
 {% for tag in collections.tags %}
 <p>
-  <a href="{{ tag.url | url }}/" >{{ tag.data.name }}</a>
+  <a href="{{ tag.url | url }}" >{{ tag.data.name }}</a>
```
beechnut, line 11: Nice catch :)

templates/footer.html
```diff
@@ -48,7 +48,7 @@ <h2 class="font-heading-md text-medium margin-top-4 tablet:margin-top-0 margin-b
         </h2>
         <ul class="usa-list usa-list--unstyled font-sans-sm list-item-spacing-2">
           {% for item in navigation.contact %}
-            <li><a href="{{ item.href | url }}">{{ item.text }}</a></li>
+            <li><a href="{{ item.href | url }}" {% if item.class %}class="{{ item.class }}"{% endif %}>{{ item.text }}</a></li>
```
beechnut, line 51: This content, template, and data are all the same shape as the social_media partial. I'm going to see if we can DRY this out.

templates/subnav.html
beechnut, line 1: clarification: Did your diff show that this gets used? I have a vague memory of deleting this because we don't render this on the 18F site, whereas it does get used in the guides.

cantsin: This is indeed used. To be fair, the footer is weird because part of it pulls from config and part is also hard coded (such as the social media links). All I wanted to do here was to add the external link icon, if we can do that another way, great!

templates/footer.html
```diff
@@ -58,30 +65,18 @@ <h2 class="font-heading-md text-medium margin-top-4 tablet:margin-top-0 margin-b
...
+              <a href="{{ item.link | url }}"
+                 class="{% if item.html.class %}{{ item.html.class }}{% endif %}"
+                 {% if item.html.rel %}rel="{{ item.html.rel }}"{% endif %}
+                 >
+                {% assign img_class = "mmaxw-205 margin-right-1 " | append: item.image.html.class %}
+                {% image_with_class item.image.dark img_class %}
+                {{- item.platform -}}
+              </a>
+            </li>
+          {% endfor %}
```
cantsin, line 79: Muuuuuuuch better.

templates/subnav.html
cantsin: Your fix LGTM and is awesome 🎉

PR #4062, opened 2024-08-15
the the to the

jskinne3:
# Pull request summary

Replace `the the` with `the`


beechnut:
This is beautiful, and it's hilarious how many of these we have.

So it's with a heavy heart I have to say — there's a content freeze on the 18F website right now while we finalize the migration and switch over to 11ty. I would normally suggest making the changes on `replatform-main`, but that'll add a little more complexity to our diff-based QA process, and the last thing we need right now is more complexity.

Recommendation: wait for the merge (🤞🏼 tomorrow?) and open a new PR with these changes then

cantsin:
OMG i love this so much. the the diff is particularly eye opening 😁

PR #4085, opened 2024-08-28
the the to the

jskinne3:
# Pull request summary

Replace `the the` with `the`

A re-do of https://github.com/18F/18f.gsa.gov/pull/4062


content/pages/projects/services/doi-nrrd.md
```diff
@@ -52,5 +52,5 @@ become a model for other countries working on opening up extractive
 industries data. The Office of Natural Resources Revenue, the Interior
 office responsible for the report, has also adopted user-centered design
 and plans to apply it to their other digital services. You can follow
-the The Natural Resources Revenue Data project and the Office of Natural
+The Natural Resources Revenue Data project and the Office of Natural
```
beechnut, line 55: double-check: should we preserve the lower-case "the" instead of "The"?

beechnut (approved):
After double-checking the the comment I left about the vs The, this looks good to merge!

Thanks thanks for doing this!

content/pages/projects/services/doi-nrrd.md
jskinne3: Thanks for checking this. I'll change it to preserving "the" because it does not appear clear that "The" is a part of the title of this thing

PR #4086, opened 2024-08-28
Update markdown-it to fix SNYK-JS-MARKDOWNIT-6483324

caleywoods:
# Pull request summary
Updates the `markdown-it` markdown rendering library to v13.x.x to address [SNYK-JS-MARKDOWNIT-6483324](https://security.snyk.io/vuln/SNYK-JS-MARKDOWNIT-6483324). 

This partially addresses #4068 but the build on that PR fails I think because of the `glob` update. There will be a separate future PR to update `glob`.

PR #4068, opened 2024-08-16
[Snyk] Fix for 2 vulnerabilities

hbillings:
snyk-top-banner

#### Snyk changed the following file(s):

- `package.json`
- `package-lock.json`

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
high severity | Infinite loop <br/>[SNYK-JS-MARKDOWNIT-6483324](https://snyk.io/vuln/SNYK-JS-MARKDOWNIT-6483324) | &nbsp;&nbsp;****&nbsp;&nbsp; 
medium severity | Missing Release of Resource after Effective Lifetime <br/>[SNYK-JS-INFLIGHT-6095116](https://snyk.io/vuln/SNYK-JS-INFLIGHT-6095116) | &nbsp;&nbsp;**417**&nbsp;&nbsp; 

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/tts/project/46092824-1e87-4db3-8472-11128cb687d3?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates)
🛠 [Adjust project settings](https://app.snyk.io/org/tts/project/46092824-1e87-4db3-8472-11128cb687d3?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://support.snyk.io/hc/en-us/articles/360003891078-Snyk-patches-to-fix-vulnerabilities)

---

**Learn how to fix vulnerabilities with free interactive lessons:**

🦉 [Learn about vulnerability in an interactive lesson of Snyk Learn.](https://learn.snyk.io/?loc&#x3D;fix-pr)

[...]

PR #4088, opened 2024-08-28
Update glob to 9.x.x and 11ty SVG sprite plugin to 2.x.x

caleywoods:
# Pull request summary
Snyk tried to update our `glob` dependency to 9.x [but the build was failing](https://github.com/18F/18f.gsa.gov/actions/runs/10423507739/job/29264902514#step:7:25). It took quite a bit of digging to find what I think was the culprit which was v1.3.1 of the Eleventy SVG sprite plugin that we use to render USWDS icons on the 18f site.

[eleventy-plugin-svg-sprite v1.3.1](https://www.npmjs.com/package/eleventy-plugin-svg-sprite/v/1.3.1?activeTab=code) does not list any version of `glob` as a dependency in its own `package.json` file but it makes use of `glob` in its `src/SVGSprite` file. The code goes on to call `util.promisify(glob)` which lines up with the error I was getting before updating the plugin version about promisifying something that is already a promise since `glob` v9+ became async.

This in combination with #4086 address the PR that Snyk opened #4068.

~# Discussion/check
I see in the console now after these changes that there's a lot of output from the sass plugin about directives. That's probably worth investigating before this is merged.~


cantsin (approved):
LGTM: this is a step forward. Dunno about the sass issues, up to you if you want to merge or investigate a bit further.

caleywoods:
Merging, I don't see the USWDS output any longer locally or in the build log here. Might be a remnant of some experiment that didn't end up needing to be a part of this PR.

PR #4136, opened 2024-11-07
Internal-ize and fix all 18f.gsa.gov links

beechnut:
# Pull request summary

This PR makes 2 changes.

First, it makes all links to 18f.gsa.gov pages internal. This makes all the internal site links checkable by the link checker, which can only check internal links. This first change made a lot of links' brokenness visible for the first time.

Next, it corrects broken links. For what I think was the majority of links, they just needed a trailing slash, or to be redirected, or to remove anchors that no longer exist even when the pages still do. But for places where the links are extremely broken — like they reference a project or 18F initiative from long ago — I added "(link removed)" so it wouldn't be _too_ confusing to future readers.


adunkman (approved):
Thank you for improving these & fixing the found broken links!

PR #4137, opened 2024-11-07
Cleaning up front matter from the last year or so

beechnut:
I noticed that the "year of launches" post has no tags — they're all commented out, even though they look quite relevant.

The tags that were commented out are written using dashes, which is the wrong tag format — they need to be space-separated. So, my guess is that they were commented out because they were causing build errors.

This commit:

- Adds back commented-out tags
- Cleans up title and excerpt formats from posts from the last year or so


adunkman (approved):
Thanks for tidying this up!

